PR #30, opened 2015-04-02
Added MultiInstanceFactory registration for Simple Injector.

henkmollema:
The registration for the new `MultiInstanceFactory` delegate was missing in the Simple Injector example. Added the registration by using `GetAllInstances(t)` and updated the registration of `IMediator` too.

### DIFF
--- a/src/MediatR.Examples.SimpleInjector/Program.cs
+++ b/src/MediatR.Examples.SimpleInjector/Program.cs
@@ -22,13 +22,14 @@
         {
             var container = new Container();
             var assemblies = GetAssemblies().ToArray();
-            container.Register<IMediator>(() => new Mediator(container.GetInstance<SingleInstanceFactory>()));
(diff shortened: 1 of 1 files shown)


henkmollema:
I had to the instance creator overload for registering the `Mediator` class as the `IMediator` implementation since it has 2 public constructors, which Simple Injector doesn't like.

The creator of Simple Injector even [states](https://simpleinjector.codeplex.com/discussions/353520) that having multiple (public) constructors is an [anti-pattern](http://stackoverflow.com/a/2470404/264697).

What is your opinion on this?

jbogard:
Before you added the MultiInstanceFactory, did the sample run successfully?

henkmollema:
Yeah, it actually works. Although I don't know why. Thought this was more consistent with the other examples and the Simple Injector API.

PR #36, opened 2015-05-07
Notifications are processed simultaneously.

monnster:
Since notifications may have multiple subscribers it's better to process it simultaneously because each of them may take much time.


jbogard:
Same as #52

PR #38, opened 2015-05-19
Added a "NoContainer" example

nicolocodev:
[Today I noticed](https://twitter.com/jbogard/status/600674030098587648) that after v1.0 (I mean, in `-pre` nuget packages) MediatR doesn't have a ServiceLocator constructor dependency and I can create a Mediator instance using two new delegates. In this new example I'm using `Activator.CreateInstance` to create the instance from the Mediator interfaces


jbogard:
Merged!

PR #40, opened 2015-06-15
Added cancellation tokens to all async APIs

khellang:
When you have async APIs it's nice to be able to cancel some work being done by handlers. Since you're breaking things (for v2) anyway, I figured I might as well pitch this PR which adds a `CancellationToken` parameter to all the async APIs.

Since optional arguments [don't version very well](http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/), I figured it's best to keep the amount to a minimum. I opted for optional arguments only on the methods that are meant to be called by the consumer, i.e. `IMediator`'s methods. Shout out if you want them changed to overloads without the tokens. A third alternative is to keep methods *with* the tokens and provide extension methods passing `default(CancellationToken)` or `CancellationToken.None` instead.

### DIFF
--- a/src/MediatR.Examples/GenericAsyncHandler.cs
+++ b/src/MediatR.Examples/GenericAsyncHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace MediatR.Examples
@@ -12,7 +13,7 @@ namespace MediatR.Examples
             _writer = writer;
         }
 
-        public async Task Handle(IAsyncNotification notification)
+        public async Task Handle(IAsyncNotification notification, CancellationToken cancellationToken)
         {
             await _writer.WriteLineAsync("Got notified also async.");
         }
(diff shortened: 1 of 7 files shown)


jbogard:
In this case, a handler will always have to implement a version that takes a cancellation token? And that parameter may or may not have a value?

I think I'd rather build out additional interfaces for handlers. "ICancellableAsyncRequest" etc. as I did for async. New overload, new interfaces.

One other benefit is the Mediator class can inspect cancellation tokens (say for notification handlers).

khellang:
Yes. Handlers will still have to implement the interface (only with an extra `CancellationToken` parameter). Since it's a value type, it'll get `default(CancellationToken)`, or `CancellationToken.None`, if no token was passed to the mediator.

It's up to you. If you want another interface for this, let me know.

ChaseFlorell:
> I think I'd rather build out additional interfaces for handlers. `ICancellableAsyncRequest` etc. as I did for async. New overload, new interfaces.

Yes, extend it with an interface.

michielrutting:
I'm also in favour of additional interfaces.

khellang:
Sent a new PR over at #44 :smile:

PR #44, opened 2015-07-02
Add CancellationTokens - Take 2

khellang:
So, this is a version with separate request, notification and handler interfaces. What do you think?

I also took the liberty to refactor two things:
 - Merge `GetHandler` and `GetNotificationHandlers` methods into single paths. This'll keep the core more DRY and help if you'd want to cache the emitted types at some point.
 - Add exception handling when resolving notification handlers and make sure it doesn't throw in the `try`-block (like it did before).

If you don't want the refactorings, I can strip them out :smile: It's probably best to review this commit by commit.


jbogard:
Separate discussion, but is it time to break out all the types into individual files?

khellang:
Yeah, I guess...

khellang:
Made an extra commit, splitting out all the files.

I also split out the private wrapper types from Mediator.cs and made them internal.
Because of naming conflicts, I had to rename them, so I appended `Wrapper` to the name. Is that ok?

PR #58, opened 2015-12-18
Added caching

jchannon:
This PR caches the results to `MakeGenericType` to improve perf.

Thought I'd send this up for you to look at before I do similar work for `GetNotificationHandlers` 

cc // @khellang

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -4,6 +4,7 @@ namespace MediatR
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.Concurrent;
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
@@ -17,6 +18,9 @@ namespace MediatR
 
         private readonly MultiInstanceFactory _multiInstanceFactory;
 
+        private readonly ConcurrentDictionary<Type, Type> _genericHandlerCache;
+        private readonly ConcurrentDictionary<Type, Type> _wrapperHandlerCache;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
         /// </summary>
@@ -26,6 +30,8 @@ namespace MediatR
         {
             _singleInstanceFactory = singleInstanceFactory;
             _multiInstanceFactory = multiInstanceFactory;
+            _genericHandlerCache = new ConcurrentDictionary<Type, Type>();
+            _wrapperHandlerCache = new ConcurrentDictionary<Type, Type>();
         }
 
         public TResponse Send<TResponse>(IRequest<TResponse> request)
@@ -108,8 +114,8 @@ namespace MediatR
         {
             var requestType = request.GetType();
 
-            var genericHandlerType = handlerType.MakeGenericType(requestType, typeof(TResponse));
-            var genericWrapperType = wrapperType.MakeGenericType(requestType, typeof(TResponse));
+            var genericHandlerType = _genericHandlerCache.GetOrAdd(requestType, handlerType.MakeGenericType(requestType, typeof(TResponse)));
+            var genericWrapperType = _wrapperHandlerCache.GetOrAdd(requestType, wrapperType.MakeGenericType(requestType, typeof(TResponse)));
 
             var handler = GetHandler(request, genericHandlerType);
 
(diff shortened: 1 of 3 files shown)


src/MediatR/Mediator.cs
```diff
@@ -17,6 +17,9 @@ public class Mediator : IMediator
 
         private readonly MultiInstanceFactory _multiInstanceFactory;
 
+        private readonly Dictionary<Type, Type> genericHandlerCache;
+        private readonly Dictionary<Type, Type> wrapperHandlerCache;
```
khellang: `ConcurrentDictionary`?

jchannon: Not available in PCL

jchannon:
I also wonder whether we should cache the result of `Activator.CreateInstance`

src/MediatR/Mediator.cs
jbogard: Is it available in some profiles?

khellang: @jbogard What's the reason for using Profile259? Would it make sense to switch to `dotnet5.1`, which essentially is Profile259, but will work for future platforms as well.

jchannon: I have no clue about profiles/PCLs so I'll have to let you decide what's best

jbogard: @khellang Just historical. I'm moving to .NET platform standard eventually, the question is whether to make a pitstop at `dotnet5.1` or not (I've had problems with it). It probably wouldn't hurt just to jump to it.

jchannon:
Added caching for `publish` and `Activator.CreateInstance`

jbogard:
I merged all the xproj stuff, can you merge/rebase?

jchannon:
Will do. Any changes required? @khellang didn't like this the last time he
looked at it

jbogard:
You'll likely be able to use the concurrent dictionary stuff now.

khellang:
Yeah, that instance cache is bad. It effectively makes all handlers singletons, regardless of container lifetime scope. It's also horrible to cache by hash code. What about compiling new-expressions and caching those lambdas instead of calling `Activator.CreateInstance`?

jbogard:
Also, one thing worth including are benchmarks about how much faster it is. Nearly all of my MediatR calls hit a DB or web service, so 1 vs 10 microseconds of reflection doesn't matter if I'm waiting 10-2000ms for an external call to finish.

jchannon:
Could use this? https://github.com/jchannon/MediatRTest/

Shall we implement @khellang idea?

khellang:
I just did a quick measurement of `Activator.CreateInstance` and compiled expressions on Windows. For default ctors, it doesn't matter, but for parameterized ctors, `Activator.CreateInstance` was about 46x slower. I ran the benchmark for 1.000.000 iterations and it completed in 774ms. At that point, I think we're talking about micro optimizations. 774ms on 1.000.000 requests is nothing.

khellang:
BTW, @jbogard, have you seen https://github.com/dotnet/roslyn/pull/6644? When/if that lands, you'd be able to bypass reflection completely. You can just add a generic parameter:

```csharp
TResponse Send<TRequest, TResponse>(TRequest request) where TRequest : IRequest<TResponse>;
```

If you did this today, you'd end up with

```csharp
mediator.Send<Request, Response>(request);
```

Buf after the PR lands, you'd be able to infer all the type arguments

```csharp
mediator.Send(request);
```

jbogard:
@khellang well crap let's just wait for that then

jchannon:
That roslyn issue has been closed

khellang:
It's still at https://github.com/dotnet/roslyn/pull/7850

jchannon:
Hi, I've come back to this and tested without instance caching but with type caching and the results are

| Runtime  | Results  |   
|---|---|---|---|---|
| Mono 3.12.1 (with type caching)  | 63ms  |  
| Mono 4.2.2 (with type caching)  | 44ms  |   
| Mono 3.12.1 (nuget)  | 98ms  |   
| Mono 4.2.2 (nuget)  | 107ms  |   

The tests is here - https://github.com/jchannon/MediatRTest/

Would you consider merging this without the instance cache?

khellang:
Compared to what without the caching?

jchannon:
Sorry, yes. Tests with my PR and without using the nuget package

Yantrio:
https://github.com/jchannon/MediatRTest/blob/master/Program.cs#L19

jchannon:
Updated table with use of concurrent dictionary

jchannon:
Have spoken to Mono team and I gave them a repro for ConcurrentDictionary and there seems to be a performance problem with it on 4.2.2 hence why we see a -32% slow down. I think its fixed in latest master and it appears to be quicker than 3.12.1.

However, hope you think the concurrentdictionary caching is useful and would love to see this merged

PR #62, opened 2016-01-04
added: DryIoc sample targeting .NET Core

dadhi:
Hello @jbogard 

I've added [DryIoc.Dnx](https://bitbucket.org/dadhi/dryioc/overview) sample. Compiles and is running just fine.


khellang:
:shipit:

PR #64, opened 2016-02-05
Removing awaiters

jbogard:
Thoughts @danielmarbach ?

### DIFF
--- a/src/MediatR/AsyncRequestHandler.cs
+++ b/src/MediatR/AsyncRequestHandler.cs
@@ -11,7 +11,7 @@ namespace MediatR
     {
         public async Task<Unit> Handle(TMessage message)
         {
-            await HandleCore(message).ConfigureAwait(false);
+            await HandleCore(message);
 
             return Unit.Value;
         }
--- a/src/MediatR/AsyncRequestHandler.cs
+++ b/src/MediatR/AsyncRequestHandler.cs
@@ -11,7 +11,7 @@ namespace MediatR
     {
         public async Task<Unit> Handle(TMessage message)
         {
-            await HandleCore(message);
+            await HandleCore(message).ConfigureAwait(false);
 
             return Unit.Value;
         }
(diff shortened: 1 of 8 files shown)


src/MediatR/AsyncRequestHandler.cs
```diff
@@ -11,7 +11,7 @@ public abstract class AsyncRequestHandler<TMessage> : IAsyncRequestHandler<TMess
     {
         public async Task<Unit> Handle(TMessage message)
         {
-            await HandleCore(message);
+            await HandleCore(message).ConfigureAwait(false);
```
WilliamBZA: Since MediatR is a library that clients will call and consume, are you not better off using ConfgureAwait(true) so the caller can decide the what the continuation context should be?

danielmarbach: @jbogard Should the users be able to access things which require context? i.e. user interface components, http context...

jbogard: What I really wanted to do was remove the async/await altogether, and return the HandleCore task result...somehow. But I was bridging the difference between `Task` and `Task<T>`.

@danielmarbach is there a way for me to convert the void task into a task that returns the Unit type that I stole from F#?

danielmarbach: In your case this is not really achievable without introducing unnecessary complexity. I'd rather go and introduce another interface on the library level like `IAsyncRequestHandler<TMessage>` and let the mediator handle the smarts

jbogard: There already is one, but it returns `Task<T>`. Stupid void returns

jbogard: And to answer your question - yes, people would want to access things with context, inside that internal HandleCore method.

danielmarbach: Then remove `ConfigureAwait(false)`

danielmarbach: @jbogard For your interest https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/ControllerActionExecutor.cs#L173

jbogard: Ah yeah then that's exactly what I'm doing.

danielmarbach:
@jbogard do you need more input?

PR #67, opened 2016-02-05
Removed CancellableAsync handlers

danielmarbach:
This is a proposal PR. I think the `CancellableAsyncRequestHandler` is just unnecessary noise. It is kinda handy to always have a `CancellationToken` in the scope of the handle method. By using `default(CancellationToken)` you get `CancellationToken.None`.

This would be a major breaking change. If you like it I rebase and add the two failing tests


jbogard:
Yeah that is a breaking change. It's probably the right one though, since users are calling the handlers from their application code directly, rather than like NSB with the framework calling it.

danielmarbach:
Should I finish it?

Do you use any kind of obsoletion techniques?

khellang:
I actually PR'd this quite a while ago (#40), but was asked to  split it out as a separate interface (#44). Are we going back on this now? If so, :+1: from me.

jbogard:
Not really for this library - I just rev to 3.0. Once merged I'd modify the project.json. 

One thing that worries me are default param values, those don't version, but I don't see them changing

danielmarbach:
@khellang history repeats itself

danielmarbach:
So what's up with this?

jbogard:
Can you rebase?

abatishchev:
@danielmarbach ping

danielmarbach:
Hopefully I can rebase after my vacation (approx. 1 week)

abatishchev:
@danielmarbach can you please rebase?

danielmarbach:
rebasing right now

danielmarbach:
Rebased

khellang:
There's a failing test; `MediatR.Tests.AsyncPublishCancelTests.Should_throw_cancelled`

danielmarbach:
Wow this was a crappy experience. Shouldly doesn't have a ThrowsAsync on .net standard so I had to trick around that  a bit

danielmarbach:
Green, green, green...

jbogard:
OK the way I'm thinking of going here with 3.0 is including the pipeline as part of the MediatR setup, which would mean I would standardize a pipeline, but not worry about the individual kinds of handlers.

Issue #94, opened 2016-08-09
Using a single event class implementing both INotification and IAsyncNotification leads to cache exception


sickboy:
```
Unable to cast object of type 'MediatR.Internal.NotificationHandlerWrapper`1[MyEvent]' to type 'MediatR.Internal.AsyncNotificationHandlerWrapper'.
```

In the meantime I found that in most of my projects usages there was a clear separation between events that are handled async and those that are handled sync, but I liked that I could just raise the events, and not worry about (future) consumer kinds. IMO it's a consumer, and not publisher concern.

jbogard:
Ah hmmmm, is this just a bug in the cache then?

sickboy:
@jbogard yea the same storage is used for all the kinds of handlers (Notification/Handler, AsyncNotification/AsyncHandler, AsyncNotificationWithCancellation/AsyncHandlerWithCancellation), based on the runtime type of the message class.

valdisiljuconoks:
Yes, stepped on the same issue recently.

maca88:
+1

PR #95, opened 2016-08-09
Add configurable cache and cache separately per TWrapper

sickboy:
Configurable cache and solution to #94. Defaults to a static cache currently.
I think it would be better to change the approach to a non static class though, as one could register a singleton cache if desired anyway.


jbogard:
Just removed all caches in favor of a cached handler factory.

PR #100, opened 2016-08-31
Add void handler interfaces

geirsagberg:
Adds `IRequestHandler<TMessage>`, `IAsyncRequestHandler<TMessage>` and `ICancellableRequestHandler<TMessage>` interfaces that do not have a response, and can be used instead of the respective base classes.


abatishchev:
I personally like this idea and prefer to return `Task` rather than `Task<Unit>`.

Issue #98, opened 2016-08-21
How to properly pipeline async handlers?


abatishchev:
Given you have a pipeline or a flow of async notification handlers, say `Pre1`, `Pre2` and `Final`, of the same notification type. How to run them one after another.

Currently I use use Simple Injector to decorate Final with Prep1 and Prep1 with Prep2:

```csharp
container.RegisterCollection<IAsyncNotificationHandler<Foo>(
  new[]
  {
    typeof(FinalHandler)
  });
container.RegisterDecorator<IAsyncNotificationHandler<Foo>, Prep1Handler>();
container.RegisterDecorator<IAsyncNotificationHandler<Foo>, Prep2Handler>();
```

Instead I propose method `PipelineAsync()` with a naïve implementation such as:
```csharp
public async Task PipelineAsync(IAsyncNotification notification)
{
  foreach(var handler in GetNotificationHandlers(notification)
  {
    await handler.Handle(notification);
  }
}
```
or add an overload to existing method `PublishAsync()` accepting `PublishOptions` with a property  of type enum:
```csharp
enum HandlerRunMode
{
  // Default = Parallel, // optionally
  Parallel,
  Sequential
}

public Task PublishAsync(IAsyncNotification notification)
{
  return PublishAsync(notification, new PublishOptions { RunMode = HandlerRunMode.Parallel });
}

public Task PublishAsync(IAsyncNotification notification, PublishOptions options)
{
  switch (options.RunMode)
  {
    case (HandlerRunMode.Parallel): return ParallelPublishAsyncImpl(notification);
    case (HandlerRunMode.Sequential): return SequentialPublishAsyncImpl(notification);
  }
}
```
Makes sense?

abatishchev:
Can you please review?

abatishchev:
@jbogard can you please take a look?

samueldjack:
I've just been bitten by this, and it's taken me the best part of 5 hours to work out what was going on :-(

In my case I had two handlers listening to the same event, and each handler used the same DbContext to make async calls to my database. I kept getting System.NotSupportedException thrown by the Entity Framework async methods with the message "A second operation started on this context before a previous asynchronous operation completed". I couldn't fathom it, because my expectation was that the handlers would be executed in sequence rather than in parallel.

abatishchev:
Moving discussion to the pull request

PR #101, opened 2016-09-01
Added option to enable PublishAsync to invoke handlers sequentially

samueldjack:
This helps to address #98 by adding a configuration option to enable invoking async notification handlers sequentially rather than in parallel.

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -21,7 +21,7 @@ namespace MediatR
         private readonly ConcurrentDictionary<Type, Type> _genericHandlerCache;
         private readonly ConcurrentDictionary<Type, Type> _wrapperHandlerCache;
 
-        private PublishAsyncOptions _publishOption;
+        private readonly PublishAsyncOptions _publishOption;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
@@ -29,7 +29,7 @@ namespace MediatR
         /// <param name="singleInstanceFactory">The single instance factory.</param>
         /// <param name="multiInstanceFactory">The multi instance factory.</param>
         public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory) :
-            this(singleInstanceFactory, multiInstanceFactory, new MediatorConfiguration())
+            this(singleInstanceFactory, multiInstanceFactory, new MediatorConfiguration() { PublishAsyncOptions = PublishAsyncOptions.Parallel})
         {
         }
 
@@ -125,8 +125,7 @@ namespace MediatR
         private Task PublishParallelAsync(IAsyncNotification notification, IEnumerable<AsyncNotificationHandlerWrapper> notificationHandlers)
         {
             var tasks = notificationHandlers
-                .Select(handler => handler.Handle(notification))
-                .ToArray();
+                .Select(handler => handler.Handle(notification));
 
             return Task.WhenAll(tasks);
         }
@@ -135,8 +134,7 @@ namespace MediatR
             IEnumerable<CancellableAsyncNotificationHandlerWrapper> notificationHandlers)
         {
             var tasks = notificationHandlers
-                .Select(handler => handler.Handle(notification, cancellationToken))
-                .ToArray();
+                .Select(handler => handler.Handle(notification, cancellationToken));
 
             return Task.WhenAll(tasks);
         }
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -21,17 +21,32 @@ namespace MediatR
         private readonly ConcurrentDictionary<Type, Type> _genericHandlerCache;
         private readonly ConcurrentDictionary<Type, Type> _wrapperHandlerCache;
 
+        private PublishAsyncOptions _publishOption;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
         /// </summary>
         /// <param name="singleInstanceFactory">The single instance factory.</param>
         /// <param name="multiInstanceFactory">The multi instance factory.</param>
-        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory)
+        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory) :
+            this(singleInstanceFactory, multiInstanceFactory, new MediatorConfiguration())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Mediator"/> class.
+        /// </summary>
+        /// <param name="singleInstanceFactory">The single instance factory.</param>
+        /// <param name="multiInstanceFactory">The multi instance factory.</param>
+        /// <param name="configuration">The configuration options</param>
+        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory, MediatorConfiguration configuration)
         {
             _singleInstanceFactory = singleInstanceFactory;
             _multiInstanceFactory = multiInstanceFactory;
             _genericHandlerCache = new ConcurrentDictionary<Type, Type>();
             _wrapperHandlerCache = new ConcurrentDictionary<Type, Type>();
+
+            _publishOption = configuration.PublishAsyncOptions;
         }
 
         public TResponse Send<TResponse>(IRequest<TResponse> request)
@@ -73,20 +88,65 @@ namespace MediatR
 
         public Task PublishAsync(IAsyncNotification notification)
         {
-            var notificationHandlers = GetNotificationHandlers(notification)
-                .Select(handler => handler.Handle(notification))
-                .ToArray();
+            var notificationHandlers = GetNotificationHandlers(notification);
 
-            return Task.WhenAll(notificationHandlers);
+            if (_publishOption == PublishAsyncOptions.Parallel)
+            {
+                return PublishParallelAsync(notification, notificationHandlers);
+            }
+            else
+            {
+                return PublishSequentialAsync(notification, notificationHandlers);
+            }
         }
 
         public Task PublishAsync(ICancellableAsyncNotification notification, CancellationToken cancellationToken)
         {
-            var notificationHandlers = GetNotificationHandlers(notification)
+            var notificationHandlers = GetNotificationHandlers(notification);
+
+            if (_publishOption == PublishAsyncOptions.Parallel)
+            {
+                return PublishParallelAsync(notification, cancellationToken, notificationHandlers);
+            }
+            else
+            {
+                return PublishSequentialAsync(notification, cancellationToken, notificationHandlers);
+            }
+        }
+
+        private async Task PublishSequentialAsync(IAsyncNotification notification, IEnumerable<AsyncNotificationHandlerWrapper> notificationHandlers)
+        {
+            foreach (var handler in notificationHandlers)
+            {
+                await handler.Handle(notification);
+            }
+        }
+
+        private Task PublishParallelAsync(IAsyncNotification notification, IEnumerable<AsyncNotificationHandlerWrapper> notificationHandlers)
+        {
+            var tasks = notificationHandlers
+                .Select(handler => handler.Handle(notification))
+                .ToArray();
+
+            return Task.WhenAll(tasks);
+        }
+
+        private Task PublishParallelAsync(ICancellableAsyncNotification notification, CancellationToken cancellationToken,
+            IEnumerable<CancellableAsyncNotificationHandlerWrapper> notificationHandlers)
+        {
+            var tasks = notificationHandlers
                 .Select(handler => handler.Handle(notification, cancellationToken))
                 .ToArray();
 
-            return Task.WhenAll(notificationHandlers);
+            return Task.WhenAll(tasks);
+        }
+
+        private async Task PublishSequentialAsync(ICancellableAsyncNotification notification, CancellationToken cancellationToken, IEnumerable<CancellableAsyncNotificationHandlerWrapper> notificationHandlers)
+        {
+            foreach (var handler in notificationHandlers)
+            {
+                await handler.Handle(notification, cancellationToken);
+            }
         }
 
         private RequestHandlerWrapper<TResponse> GetHandler<TResponse>(IRequest<TResponse> request)
--- a/src/MediatR/MediatorConfiguration.cs
+++ b/src/MediatR/MediatorConfiguration.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace MediatR
+{
+    public enum PublishAsyncOptions
+    {
+        Parallel,
+        Sequential
+    }
+
+    public class MediatorConfiguration
+    {
+        public PublishAsyncOptions PublishAsyncOptions { get; set; } = PublishAsyncOptions.Parallel;
+    }
+}
--- a/test/MediatR.Tests/AsyncPublishTests.cs
+++ b/test/MediatR.Tests/AsyncPublishTests.cs
@@ -75,5 +75,37 @@
             result.ShouldContain("Ping Pong");
             result.ShouldContain("Ping Pung");
         }
+
+        [Fact]
+        public async Task Should_invoke_handlers_in_order_when_Sequential_Publishing_is_chosen()
+        {
+            var builder = new StringBuilder();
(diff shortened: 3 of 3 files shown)


src/MediatR/Mediator.cs
```diff
@@ -73,20 +88,65 @@ public void Publish(INotification notification)
...
         }
 
-        public Task PublishAsync(ICancellableAsyncNotification notification, CancellationToken cancellationToken)
+        private Task PublishParallelAsync(ICancellableAsyncNotification notification, CancellationToken cancellationToken,
+            IEnumerable<CancellableAsyncNotificationHandlerWrapper> notificationHandlers)
         {
-            var notificationHandlers = GetNotificationHandlers(notification)
+            var tasks = notificationHandlers
                 .Select(handler => handler.Handle(notification, cancellationToken))
                 .ToArray();
```
abatishchev: Please drop `ToArray()` (here and elsewhere). This is not needed and actually can lead to a wrong behavior. Send the IEnumerable directly to `Task.WhenAll()` and it will do the needful to run them in parallel properly.

src/MediatR/Mediator.cs
```diff
@@ -73,20 +88,65 @@ public void Publish(INotification notification)
 
         public Task PublishAsync(IAsyncNotification notification)
         {
```
abatishchev: To not introduce breaking change, I'd make this an overload having the default behavior.

src/MediatR/Mediator.cs
```diff
@@ -21,17 +21,32 @@ public class Mediator : IMediator
...
 
+        private PublishAsyncOptions _publishOption;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
         /// </summary>
         /// <param name="singleInstanceFactory">The single instance factory.</param>
         /// <param name="multiInstanceFactory">The multi instance factory.</param>
-        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory)
+        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory) :
```
abatishchev: Adding one more ctor is a breaking chance since some DI containers (such as Simple Injector) will stop working. So I'd keep only one and rely on the options parameters sent to an overload accepting it.

abatishchev:
hey @samueldjack, thanks for converting my question to a pr!

src/MediatR/Mediator.cs
khellang: > This is not needed and actually can lead to a wrong behavior.

This is false. [The source](http://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,69351c6da968e5d1) clearly shows that both `WhenAll(IEnumerable<Task>)` and `WhenAll(Task[])` ends up in the same [`InternalWhenAll`](http://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,2529f60968787540) method after checking for `null` tasks.

> Send the IEnumerable directly to Task.WhenAll() and it will do the needful to run them in parallel properly.

As mentioned above, it'll do the exact same thing as doing `ToArray()` before passing it in. It doesn't matter.

abatishchev: False or true, anyway let's be more friendly, right? I'm basing my statement on my experience that calling ToArray() kick-offs these tasks before they're reaching Task.WhenAll() what doesn't help much during debugging. Thanks for pointing out to the source code. But still I'd pass IEnumerable directly and let TPL do its work.

khellang: > False or true, anyway let's be more friendly, right?

I'm sorry. It wasn't my intention to be unfriendly :confused: I just wanted to point out that what you said is incorrect and based on false assumptions. @samueldjack did nothing wrong by calling `ToArray` before passing the tasks to `WhenAll`, or simply leaving the code as-is. The sooner the tasks are  "kicked off", the better, IMO :smile:

> I'm basing my statement on my experience that calling ToArray() kick-offs these tasks before they're reaching Task.WhenAll() what doesn't help much during debugging.

Yes, the enumerable is enumerated when calling ´ToArray´, but does it matter if that happens before the call to `WhenAll`? It's only a matter of time before `ToArray` [is called anyway](http://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,6016)... If anything, you could argue that passing an array is marginally *faster* than calling the enumerable overload, because of fewer checks. I'm not sure how this affects debugging?

abatishchev: Maybe just a matter of taste or semantic perspective but since it's literally a pass-thru calll I'd keep it a pass-thru call. Agree that materialising a hot enumerable can have side effects? Even it will happen anyway, and the source code is available, I would rely on the behavior of BCL/TPL rather than introduce possible side effects in MediatR. It's works very well as a 'glue' between application layers because it's lightweight. And I'd love to see it not to change.

samueldjack: My philosophy was to keep existing code the way it was as much as possible: the original PublishAsync used ToArray() so I just preserved that.

Having said that, I think call stacks during debugging might become somewhat easier to understand if the task is initiated with the WhenAll rather than before, so I'd have a preference for dropping ToArray too.

src/MediatR/Mediator.cs
samueldjack: The problem with this is that I'd then have to update every call site in my code to get the behaviour I want - since I want sequential behaviour whenever events are published.

If the default option on the Mediator is configured correctly then there would be no breaking change in behaviour irrespective of how the method is called.

src/MediatR/Mediator.cs
samueldjack: I hadn't thought about the impact of adding an extra constructor on DI resolution. 

Like I said in my comment on overloads of PublishAsync, I'd like to make this publish behaviour a global option so that I don't have to update every call site where events are published.

I've just checked out the [breaking change rules for CoreFx](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md) and by my reading adding an extra constructor wouldn't be considered a breaking change.

However, if the extra constructor issue is serious, perhaps we could have a property on Mediator to configure the default behaviour?

src/MediatR/Mediator.cs
abatishchev: Both points are valid, sound good to me!

src/MediatR/Mediator.cs
abatishchev: I see. This would be an issue for me as well, will have to update every call site. So settings behavior in one place is a good idea.

src/MediatR/Mediator.cs
abatishchev: I'd rather go ahead with this change via ctor than add a property.
Either way it should be acknowledged/highlighted, and major version be bumped.
@jbogard your thoughts?

src/MediatR/Mediator.cs
```diff
@@ -21,17 +21,32 @@ public class Mediator : IMediator
         private readonly ConcurrentDictionary<Type, Type> _genericHandlerCache;
         private readonly ConcurrentDictionary<Type, Type> _wrapperHandlerCache;
 
+        private PublishAsyncOptions _publishOption;
```
abatishchev: Until a property is added, it can be made `readonly``.

src/MediatR/Mediator.cs
```diff
@@ -21,17 +21,32 @@ public class Mediator : IMediator
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
         /// </summary>
         /// <param name="singleInstanceFactory">The single instance factory.</param>
         /// <param name="multiInstanceFactory">The multi instance factory.</param>
-        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory)
+        public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactory multiInstanceFactory) :
+            this(singleInstanceFactory, multiInstanceFactory, new MediatorConfiguration())
```
abatishchev: I'd make it more readable: `new MediatorConfiguration { PublishAsyncOptions = PublishAsyncOptions.Parallel  }`, otherwise reader needs to go to the class's definition and learn the behavior from there

abatishchev:
@jbogard any chance to review please?

dotnetjunkie:
@abatishchev

> Adding one more ctor is a breaking chance since some DI containers (such as Simple Injector) will stop working

This shouldn't be a problem, since you shouldn't use your container's auto-wiring capabilities to create types of external libraries and frameworks, because of exactly this. Instead register external types using a delegate. This way you have full control over the constructor that is called.

Note that this is not a Simple Injector specific thing. This advice holds for all containers. Every container has its own unique way to select constructors.

abatishchev:
Hi Steven! I don't say it's a problem with SI or there are no workarounds. But if one already uses their container's  auto-wiring capability than it will be a breaking change for them.
I'm not quite conveinced why one should not use it if it works and works flawlessly. In particular, MediatR has single ctor from the very beginning so why don't write `Register<IMediator, Mediator>()` rather than `Register<IMediator>(() => new Mediator(...))` when you can? When you can't then you can't, and obviously have no other options. But until that it's nice to have such a choice.
That being said, is there workaround or other (and better)  way to achieve the same? Yes. Would it be a breaking change? Yes. Would it be a big deal? No.

dotnetjunkie:
>But if one already uses their container's auto-wiring capability than it will be a breaking change for them.

Be careful calling this a breaking change, because this would limit a reusable library builder like Jimmy to make changes to its library that would usually be stated as 'non-breaking' by [The Framework Design Guidelines](https://www.amazon.com/Framework-Design-Guidelines-Conventions-Libraries/dp/0321545613).

Adding overloaded constructors is typically the way for a framework builder to make these changes, because the signature of a constructor can't be changed (since that already is a breaking change).

> I don't say it's a problem with SI

I would even say that this is less the problem with Simple Injector. In the case of Simple Injector you get a very clear exception message, which allows you to change your configuration. With a container that does constructor resolution, the framework builder never really knows which constructor will be selected (since each DI container has other rules for selecting constructors). And when you upgrade your application to the newest version of that external library, everything might keep compiling and no error is thrown when the application runs, but the application could still function incorrectly (but without warning or error), because the wrong constructor has be selected by your container. 

If Jimmy follows the FDG (and I know he does), he should be able to add constructor overloads. If this breaks your application, it's not Jimmy's fault; it's yours.

> > so why don't write Register<IMediator, Mediator>() rather than Register<IMediator>(() => new Mediator(...)) when you can?

It's important to make the distinction between types you own and control and for which you know they have just [a single constructor](https://www.cuttingedge.it/blogs/steven/pivot/entry.php?id=97), and types of external libraries where you don't have any control over the number of constructors they have. In that case, my advice is to always call the constructor using C# code, not through reflection. Calling such constructor with normal C# code should not cause any maintenance problems, because their constructor signature will never change, compared to the types you control. We typically see constructor signatures change regularly while developing an application.

Of course I must admit here that MediatR is not your typical external library, so depending on how we see MediatR, the above description might not completely hold for MediatR.

When we follow the SOLID principles, they guide us to hiding external libraries from our application code; we create application-specific abstractions and create adapters to hide the external libraries behind. This however does not work with MediatR, since its goal is to provide you with the core abstractions to build your application around. So in this sense, MediatR must perhaps be seen as the central part of our application, otherwise we would be breaking SOLID big time. 

On the other hand, MediatR is a reusable library used in hundreds (or perhaps even thousands) of applications and meant for general use. This immediately causes conflict with it being a central part of the application, since its changes and updates are out of control of the application, and its types must be treated differently than other application types (because of the possible additions of constructors for instance).

So whether or not you use auto-wiring to build up MediatR types depends on where you place MediatR in your application architecture, but do keep in mind that you don’t control MediatR.

abatishchev:
Indeed, these all make sense, even somewhere I could argue, but I'm a way worse speaker than you, so I have no other options than admit your rightness :)
The only thing I meant saying about braking change was not against such change but to bump tye major version.

jbogard:
So....I don't like this for the main reason that it's a very specific. I intentionally try to leave behavior out of the mix in the Mediator class, because it's opaque, you can't see what it does.

What I'd rather do is allow you to customize the behavior in any way you like, but not through a switch/if-then. Even if this is just a thing where we structure things so that you can subclass and override a virtual method, I'd rather do that than such a specific configuration here.

abatishchev:
So maybe for a separate behavior to have a separate method? Let PublishAsync() run in parallel and something else (e.g. PipelineAsync() or RunSequnceAsync()) run sequentially?

But still passing an options parameter where one specifies the behavior is quite visible imo.

jbogard:
No - no options parameter. If something was going to be passed in, I'd rather it be a separate interface. INotificationPublisher or similar. No if/switch statement, make it a strategy pattern.

But overall, I'd rather subclass-and-override first. At least give people the option to add their own behaviors first before having a breaking change like this.

abatishchev:
Accepting `INotificationPublisher` makes sense. I'd ship MediatR with 2 implementations: Sequential and Parallel.

Or indeed subclass-and-override. Inherit Mediator by SequentialMediator and override PublishAsync, or decorate Mediator with SequentialMediatorDecoratory, proxy all methods but PublishAsync. Is this something you would ship with MediatR by default? If yes, I would submit a separate pull request.

abatishchev:
Please check [this draft](https://github.com/jbogard/MediatR/compare/master...abatishchev:pubseq-1?expand=1) out: an `PublishAsync()` overload accepting `IAsyncNotificationPublisher` with its default implementation. Drawbacks: the implementation is internal since depends on internal wrapper class.

Here's [another draft](https://github.com/abatishchev/MediatR/tree/seqdec-1): a sequential decorator. Drawbacks: had to extract GetNotifications() method into few internal helpers to avoid code duplication.

abatishchev:
The task would be simplified once https://github.com/jbogard/MediatR/pull/67 is merged.

abatishchev:
Also if make cache static (and internal) as per https://github.com/jbogard/MediatR/pull/73 the task would be simplified even more.

abatishchev:
I published one more draft: this time `[SequentialMediator](https://github.com/abatishchev/MediatR/blob/seqdec-2/src/MediatR/SequentialMediator.cs)` inherits `Mediator` and [overrides ](https://github.com/abatishchev/MediatR/blob/seqdec-2/src/MediatR/Mediator.cs#L77)`PublishAsync` in a manner that no internal classes have to be made public.

samueldjack:
Having thought about this further, I'm inclined to lean away from the sub-classing approach towards injecting a behaviour into the mediator. The biggest factor for me is that the subclassing approach immediately limits composability. It only affects one aspect of the Mediators behaviour (async behaviour) and yet it layers on top of the whole class.

Imagine another scenario where the sync behaviour needed to be customised and a subclass was introduced to support that. Then imagine wanting to support both the custom sync behaviour and the custom async behaviour. The only way to do that would be to reimplement the one behaviour as a further subclass of the other.

I guess the issue here is that the Mediator is a facade, pulling a couple of different kinds of behaviour (sync and async publishing) into one convenient interface. Thus the design needs to allow for independent variations in how the behaviours behind the facade are implemented.

abatishchev:
@samueldjack sorry, missed to respond in time, I completely agree!

jbogard:
OK so this is all getting changed in 3.0. Stay tuned.

abatishchev:
Cool!

jblackburn21:
I can work on a PR to v3 if a direction is decided.  This would now be need for async and cancellable async handlers.  

Do you want to got the `IAsyncNotificationPublisher` route, and default to the parallel implementation?

jbogard:
Any direction that does not involve adding new constructor arguments. That'd be a breaking change and I'd need to go to 4.0.

jbogard:
That's why I suggested just a protected virtual method. It enables you to do whatever you want.

codearoo:
Hi...  I was just looking at similar topics. I found where I can create a IAsyncNotificationHandler handler which is nice to allow the handler to decide if it should block the flow.. but is it in the works still to also allow the publishing code to determine if it cares if subsequent handlers are async or not? The Publish() method (using ver 3.0.0) is blocking until all non-async handlers finish.

abatishchev:
If you're looking for a fire-and-forget behavior you can decorate your handler with one that would enqueue on the thread pool inner handler's Handle() in one or another way, e.g.:
```csharp
public FireAndForgetCommandHandlerDecorator : IAsyncNotificationHandler<Command>
{
    private readonly IAsyncNotificationHandler<Command> _inner;

    public FireAndForgetCommandHandlerDecorator(IAsyncNotificationHandler<Command> inner)
    {
        _inner = inner;
    }

    public async Task Handle(Command command)
    {
        await Task.Run(async () => await _inner.Handle(command));
   }
}
```
(I don't guarantee this code will work as described but hope you got the idea)

PR #114, opened 2016-12-12
Pipelines

jbogard:


jbogard:
@khellang it's a simple implementation but a start

src/MediatR/IPipelineBehavior.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace MediatR
+{
+    using System;
+    using System.Threading.Tasks;
+
+    public delegate Task<object> RequestHandlerDelegate();
+
+    public interface IPipelineBehavior
```
khellang: Shouldn't we make these generic? In all cases, we should be able to know `TResponse` at compile-time...

src/MediatR/Mediator.cs
```diff
@@ -37,48 +37,78 @@ public TResponse Send<TResponse>(IRequest<TResponse> request)
...
 
         public void Send(IRequest request)
         {
             var handler = GetHandler(request);
 
-            handler.Handle(request);
+            var pipeline = CreatePipeline(request, () =>
+            {
+                handler.Handle(request);
+                return Task.FromResult((object) null);
```
khellang: Is `Task.CompletedTask` available?

src/MediatR/Mediator.cs
```diff
@@ -226,6 +256,18 @@ private IEnumerable<TWrapper> GetNotificationHandlers<TWrapper>(object notificat
                 .ToList();
         }
 
+        private Task<object> CreatePipeline(object request, RequestHandlerDelegate invokeHandler)
+        {
+            var behaviors = _multiInstanceFactory(typeof(IPipelineBehavior))
```
jdaigle: This is really simple and clever. Almost _too_ simple? Have you considered a use case where one may want to select pipeline behaviors, such as based on the Type/Assembly of the request/handler?

But maybe it's as simple as putting that logic in the pipeline behavior's implementation as needed, since it has access to the request and response?

jbogard: Yeah, I would put it in the pipeline. Although if I made the pipeline generic, then the container could pick the correct instances.

src/MediatR/IPipelineBehavior.cs
jbogard: I'll open a new PR with this. Making them non-generic greatly simplifies things tho.

Issue #102, opened 2016-09-15
MediatR 3.0 ideas


jbogard:
We've been using MediatR now for 3 years in production, and I wanted to gather some lessons learned.
1. We generally create our own MediatR pipeline (see https://lostechies.com/jimmybogard/2014/09/09/tackling-cross-cutting-concerns-with-a-mediator-pipeline/)
2. While we generally have a pipeline, what goes in the pipeline is pretty different. We have validation, logging, pre-post processing - it's all over the place.
3. Registering delegates in the container is annoying and weird
4. People seem to want to customize dispatching behaviors. See #101 
5. Creating multiple mediator pipelines for the different flavors of the handler interfaces is also annoying

Any other lessons learned from folks out there? @khellang or anyone else?

dcomartin:
Often want to send/publish messages to queue/topic, but use the same MediatR API for handling the dispatching and executing once it's handed back to the process. Basically a distinction between immediately incoming in-process and having Send/Publish delegate to something else.

jbogard:
@dcomartin what does that look like for you today?

dcomartin:
Creating extension methods off IMediator.

http://codeopinion.com/background-commands-mediatr-hangfire/
https://github.com/dcomartin/Hangfire.MediatR.Demo

flytzen:
Re the pipeline... It would be very neat if we could have a standard way to define a pipeline in a similar way to how you'd build an OWIN pipeline...

jbogard:
@flytzen what does your pipeline look like today? I'm gathering examples

henkmollema:
What about async requests/notifications and handlers only? I find myself using `IAsyncRequest` all the time and use `Task.FromResult()` in a few cases where (yet) no async calls are made. It would clean up the API surface a bit.

abatishchev:
It depends on the ratio/your usage. If you have 2 sync handlers out of 50 async then yes, I would use Task.FromResult() too. But if one has mostly or only sync handlers it's unlikely the overhead they are willing and have to pay for.

abatishchev:
Or split Mediator into 2: AsyncMediator implementing IAsyncMediator (or just IMediator) and SyncMediator implementing ISyncMediator (or just IMediator).

no1melman:
@flytzen @jbogard I have integrated mediatr directly into owin, so bypassing controllers completely. It would be interesting to see more around this direct integration. Why use controllers when the only implementation is a call to Mediatr

[...]

PR #115, opened 2016-12-12
Generic pipeline

jbogard:
@khellang this doesn't actually work in StructureMap, those tests fail. I think this might introduce more complexity. But maybe we just say we only support open generics.

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -6,6 +6,7 @@ namespace MediatR
     using System.Collections.Generic;
     using System.Collections.Concurrent;
     using System.Linq;
+    using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
 
@@ -19,6 +20,7 @@ namespace MediatR
 
         private readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, Type>> _genericHandlerCache;
         private readonly ConcurrentDictionary<Type, ConcurrentDictionary<Type, Type>> _wrapperHandlerCache;
+        private static readonly MethodInfo CreatePipelineMethod = typeof(Mediator).GetTypeInfo().DeclaredMethods.Single(m => m.Name == nameof(CreatePipeline));
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Mediator"/> class.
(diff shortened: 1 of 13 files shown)


jbogard:
Looks like open generics work just fine, but you'll need to use constraints for individuals. No idea if this is supported in other containers, @khellang can Scrutor handle an enumerable of open generics?

jdaigle:
@jbogard it works in Autofac (with an almost identical registration):

    builder.RegisterGeneric(typeof(CommandHandlerPipeline<,>))
           .As(typeof(IPipelineBehavior<,>));

It also supports mixing both open and closed generic pipelines. Disappointing that it does not work in StructureMap.

jbogard:
@jdaigle meh, it would be really weird if your pipeline varied that much. Checking our existing pipelines, we don't vary them per message. Only things like request mutators, but those aren't a pipeline step in and of themselves. I do like putting in some stock pipeline behaviors though.

src/MediatR/Mediator.cs
```diff
@@ -37,48 +39,76 @@ public TResponse Send<TResponse>(IRequest<TResponse> request)
         {
             var defaultHandler = GetHandler(request);
 
-            var result = defaultHandler.Handle(request);
+            var pipeline = GetPipeline(request, () => Task.FromResult(defaultHandler.Handle(request)));
 
-            return result;
+            return pipeline.Result;
```
khellang: Deadlock?

src/MediatR/Mediator.cs
```diff
@@ -37,48 +39,76 @@ public TResponse Send<TResponse>(IRequest<TResponse> request)
...
-            handler.Handle(request);
+            RequestHandlerDelegate<Unit> invokeHandler = () =>
+            {
+                handler.Handle(request);
+                return Task.FromResult(Unit.Value);
+            };
+
+            var pipeline = GetPipeline(request, invokeHandler);
+
+            pipeline.Wait();
```
khellang: Deadlock?

src/MediatR/Mediator.cs
```diff
@@ -226,6 +256,25 @@ private IEnumerable<TWrapper> GetNotificationHandlers<TWrapper>(object notificat
                 .ToList();
         }
 
+        private Task<TResponse> GetPipeline<TResponse>(object request, RequestHandlerDelegate<TResponse> invokeHandler)
+        {
+            var requestType = request.GetType();
+            var method = CreatePipelineMethod.MakeGenericMethod(requestType, typeof(TResponse));
```
khellang: Cache?

jbogard: I'm not sure any cache matters. With many containers (including the ASP.NET Core one), you have to register the Mediator class as scoped (see https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection/blob/master/src/MediatR.Extensions.Microsoft.DependencyInjection/ServiceCollectionExtensions.cs#L188). So tbqh I'd be up for removing all the caches, it's just cognitive overhead.

src/MediatR/Mediator.cs
khellang: http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

src/MediatR/Mediator.cs
khellang: It's not about caching instances. It's about caching the generic types and methods. Those caches should be static, so it doesn't matter how the mediator is registered 😄

src/MediatR/Mediator.cs
abatishchev: Don't block this way, sooner or later will cause a deadlock. I'd use `Task.Run(() => ...).Result` instead

src/MediatR/Mediator.cs
abatishchev: I'd keep cache and make it static indeed

src/MediatR/Mediator.cs
khellang: Yep. Async over sync is fiiine. Sync over async, not so much :cry: Hence [my comment](https://github.com/jbogard/MediatR/issues/102#issuecomment-266003071) :wink:

src/MediatR/Mediator.cs
NinoFloris: My first thought was to pass a state machine into the pipeline which would be executed on every step. This would work and flip all the closed-over instances into the state machine which is passed as an input argument but this call chain pollutes the stack trace quite a bit.

If you really want to do this right you have to do some code gen with expression trees.
Had a call with Daniel (Marbach) this monday where we concluded this. As he has created something very similar for NServiceBus.

khellang: The problem with a compiled expression-based pipeline is that you lose the ability to inject services into your decorators/handlers' ctors. You have to know, at compile-time, what arguments the pipeline takes. That's why you typically end up with a `PipelineContext`, which essentially is a bag of objects, like the OWIN dictionary, that's used to flow stuff through the pipeline.

My question is; are we willing to drop the nice dependency injection in decorators/handlers for this? Has anyone done any benchmarking to see if this is needed? I don't think there's been a heap of complaints on MediatR perf to date. Has this changed?

NinoFloris: That's not accurate at all, as what NServiceBus does can very easily be adapted to using a different instance every time, neatly supplied via the input arguments. This way you remove the state machine stack trace overhead but you still have all the freedom to input new instances on every call.

It is exactly that problem what I have been trying to solve in a performant manner, and it's very much possible.

The problem here is that you want to create a dynamic call chain `next -> next -> next` of length N, only way to do that is either code gen or allocate func's.

Problem with func is that you have to close over your current handler instance, which in effect makes the full pipeline 'single use'. 

The other option you have is pulling this instance out of the func and let the func execute a method on its input argument. Say `stateMachine.DoNext(next);` which would then internally advance one step and do correct async exception bubbling etc. I've seen an example of this and it's not pretty, lot's of work to get it right and in the end you still have a polluted stack trace. 

Your third option could be to supply a `ref int index` (only possible with a delegate) and an array of pipeline steps, so each func can pick the step at that index, increment the index and run the step. However then you need to have a way to carry the ref index with you when you call the step (as it has to call next) and ref params can be finicky.

When you look at all that work you could conclude, let's just code gen these delegate body's with the index number burned into them (pseudo code): 

```cs
public Task<TResponse> DelegateOne(IPipelineBehavior[] behaviors, request, next){
    return behaviors[0].Handle(request, next);
}

public Task<TResponse> DelegateTwo(IPipelineBehavior[] behaviors, request, next){
    return behaviors[1].Handle(request, next);
}

etc
``` 

Just cache these individually, then cache pipelines based on step count and you are good to go.

jbogard:
Yeah, I did the same with AutoMapper. I'd like to get the basic behavior
right first, then convert to expression trees at some point.

PR #117, opened 2016-12-19
Collapse IMediator into one interface

jbogard:
No caching of results - yet. I'd build a cache for an action, not a type though.


jbogard:
@khellang i didn't get rid of the other interfaces :)

samples/MediatR.Examples/PingAsyncHandler.cs
```diff
@@ -4,9 +4,9 @@
 
     public class PingAsyncHandler : IAsyncRequestHandler<PingAsync, Pong>
     {
-        public async Task<Pong> Handle(PingAsync message)
+        public Task<Pong> Handle(PingAsync message)
         {
-            return await Task.Factory.StartNew(() => new Pong { Message = message.Message + " Pong" });
+            return Task.Factory.StartNew(() => new Pong { Message = message.Message + " Pong" });
```
khellang: `Task.FromResult`?

src/MediatR/Mediator.cs
```diff
@@ -31,193 +30,122 @@ public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactor
...
-            var pipeline = GetPipeline(request, invokeHandler);
-
-            return pipeline;
+                var wrapper =
+                    (RequestHandlerWrapper)
+                    Activator.CreateInstance(typeof(RequestHandlerWrapperImpl<>).MakeGenericType(requestType), handler);
+                return () =>
+                {
+                    wrapper.Handle(request);
+                    return Task.FromResult(Unit.Value);
```
khellang: Can we cache this? `Unit.Task`?

src/MediatR/Mediator.cs
```diff
@@ -31,193 +30,122 @@ public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactor
...
-
-        private TWrapper GetHandler<TWrapper, TResponse>(object request, Type handlerType, Type wrapperType)
-        {
-            var requestType = request.GetType();
-
-            var genericHandlerType = _genericHandlerCache.GetOrAdd(handlerType, new ConcurrentDictionary<Type, Type>())
-                .GetOrAdd(requestType, handlerType, (type, root) => root.MakeGenericType(type, typeof(TResponse)));
-            var genericWrapperType = _wrapperHandlerCache.GetOrAdd(wrapperType, new ConcurrentDictionary<Type, Type>())
-                .GetOrAdd(requestType, wrapperType, (type, root) => root.MakeGenericType(type, typeof(TResponse)));
+                .Select(handler => Task.Run(() => handler.Handle(notification), cancellationToken));
```
khellang: Was this `Task.Run` intentional? I think we should steer as far away from `Task.Run` as possible.

src/MediatR/Mediator.cs
jbogard: Good call.

src/MediatR/Mediator.cs
jbogard: Another option is to execute them synchronously I guess? That's probably better, but I needed a hot task.

khellang: `Task.FromResult` is always the best option to wrap a synchronous result in a `Task` :smile:

khellang: Or even better; `ValueTask`, but not sure how adoptable that is yet...

PR #134, opened 2017-02-07
Expose exceptions that occurred while resolving handlers

mycroes:
This pull requests adds caching of the exceptions that occurred while resolving handlers. If no handler was resolved, the exceptions are used as `innerException` for the exception already thrown.

Honestly I'm not sure if this is desired behavior, personally I'd favor that the `singleInstanceFactory` should only exception on issues during resolve and return `null` if no handler was registered. In that case this commit is total nonsense and only the try/catch in `GetHandler(...)` should be removed.

The pull request also features a small fix to `GetHandler(...)` being applied twice.

### DIFF
--- a/src/MediatR/Internal/RequestHandler.cs
+++ b/src/MediatR/Internal/RequestHandler.cs
@@ -1,3 +1,5 @@
+using System.Collections.ObjectModel;
+
 namespace MediatR.Internal
 {
     using System;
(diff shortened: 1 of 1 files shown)


adamhathcock:
I'm running into this myself.  It seems the exceptions just are eaten and my app gets frozen.

jbogard:
Does this fix the issue of handlers getting constructed twice?

Issue #137, opened 2017-02-12
Problem with using Wait() method in ASP.NET MVC.


sq735:
The issue is clearly illustrated by this code:

```
public class TestCommand: IRequest
{ }
```

```
public class TestHandler : IAsyncRequestHandler<TestCommand>
{
    public Task Handle(TestCommand message)
    {
        return Task.Factory.StartNew(() =>
        {
            throw new ApplicationException();
        });
    }
}
```

```
public class HomeController : Controller
{
    private readonly IMediator _mediator;

    public HomeController(IMediator mediator)
    {
        _mediator = mediator;
    }

    public  ActionResult Index()
    {
        try
        {
            _mediator.Send(new TestCommand()).Wait();  // deadlock 
        }
        catch (Exception e)
        {
            // unreachable block 
        }

        return View();
    }
}

```

If i use `await` operator `catch` block is reachable: 

 ```
try
{
    await _mediator.Send(new TestCommand());
}
catch (Exception e)
{
     // OK
}
```

but when use `Wait()` method  i have a problem, because `catch` block is unreachable.

 ```
try
{
    _mediator.Send(new TestCommand()).Wait(); // deadlock 
}
catch (Exception e)
{
     // unreachable block 
}
```

adamhathcock:
Why are you waiting at all?  Make your controller method async.

This isn't an issue specific to MediatR.

sq735:
Yes, of course, you are right, i can make controller method `async`. But i want to understand why it doesn't work when i call `Wait()` method.

For example, if i use request/response scenarios with `Wait()` method, code works well.

```
public class Result
{  }

public class TestCommandWithResult : IRequest<Result>
{  }

public class TestHandlerWithResult : IAsyncRequestHandler<TestCommandWithResult, Result>
{
	public Task<Result> Handle(TestCommandWithResult message)
	{
		return Task.Factory.StartNew(() =>
		{
			throw new ApplicationException();
			return new Result();
		});
	}
}

public  ActionResult Index()
{
        try
        {
            _mediator.Send(new TestCommandWithResult()).Wait();
        }
        catch (Exception e)
        {
            // OK 
        }

        return View();
}
```

sq735:
Perhaps I found a lines of code which can cause problems, for example the deadlock. I think we needs to use  `ConfigureAwait(false)`.

[...]

PR #142, opened 2017-02-15
Apply ConfigureAwait(false) on all awaits

mycroes:
Pull request for #137

See https://msdn.microsoft.com/en-us/magazine/jj991977.aspx for
information regarding async/await. Pipeline behavior that was expecting
to run on the SynchronizationContext that the IMediator call was made
from will now have to take care of the switch back to the original
SynchronizationContext. Some examples of doing that can be found at
https://blogs.msdn.microsoft.com/pfxteam/2011/01/13/await-anything/.


jbogard:
Fixes #137

Issue #136, opened 2017-02-10
Handler gets instantiated twice


P14x:
Hi,

I'm using Autofac and see the constructor of my Handler being executed twice.
The Autofac sample in this repo shows the same behaviour.

I could track this to the RequestHandlerImpl class.
With the changes as in this [gist](https://gist.github.com/P14x/55d9e91e87899103dfa9cd797617a71c)
(lines 38,44,49,55,56,61,67,68) this behaviour disappeared.

Do these changes make sense?
Thanks

adamhathcock:
I see the same thing with AutoFac.  Haven't investigated further on my end.

P14x:
Same issue with Ninject and Simple Injector, but StructureMap and Windsor work fine.

jbogard:
This only happens the first time you instantiate a handler, as MediatR needs to "check" to see if a handler can be created or not. But I don't cache that handler instance as you do in the gist as that means that the instance you created will be used every single time.

P14x:
Hi Jimmy, thanks for your feedback.

I see the double instantiation also for each subsequent request for a handler instance in my ASP.NET MVC application and also in the Autofac sample.

If I understand you correctly, a first call to Runner.Run as here below would do a double instantiation for the check, but the second Runner.Run call would result in only one handler constructor call.

With Autofac, there's always a double instantiation.


        public static void Main(string[] args)
        {
            var mediator = BuildMediator();

            Runner.Run(mediator, Console.Out).Wait();
            Runner.Run(mediator, Console.Out).Wait();

            Console.ReadKey();
        }

P14x:
Passing a Boolean instance field to LazyInitializer.EnsureInitialized instead of a local function variable as in this [gist](https://gist.github.com/P14x/ac14f308087da7c100419e3b89a02f9c) results into the behaviour you explain; a double instantiation once, a single one for subsequent calls.

plioi:
FWIW, I can reproduce the double instantiation as well, on all web requests rather than only the first one.

In each web request, the two constructions follow this pattern:

During the first construction, my call stack includes a line, "System.Threading.LazyInitializer.EnsureInitializedCore...".
During the second 'real' construction, the call stack does not have that line.

Andrew-Brad:
[...]

Issue #126, opened 2017-01-08
GetHandlerFactory (from RequestHandler) is not cached


remcoros:
When working on a fix for #124 I noticed the following in https://github.com/jbogard/MediatR/blob/master/src/MediatR/Internal/RequestHandler.cs#L63

Because 'initialized' is a local variable (and false), GetHandlerFactory is always called and _handlerFactory  re-instantiated. Basically making this dead code.

It is probably intended to be a cache, and the 'initialized' should be moved to a private field.

But... this introduces a possible problem with the (static) handler cache. Currently, RequestHandlerImpl is cached in a static dictionary. So when the handler factory is also cached. I loose the ability to dynamically re-bind handlers in the container during runtime.

For example, I have RequestHandlerFor&lt;Ping&gt; and I want to re-bind (at runtime) it to AsyncRequestHandlerFor&lt;Ping&gt;.

At the moment, this works, but that is a side-effect of the handler factory cache not working.

You can test/debug this by setting a breakpoint on GetHandlerFactory and change SendTests to call 'Send' multiple times.

Possible solutions:
- Make the handler factory cache work by moving 'initialized' to a private field and make the ConcurrentDictionary caches in Mediatr.cs non-static, so that we can re-initialize a new Mediatr instance in our container when re-binding handlers. https://github.com/jbogard/MediatR/compare/master...remcoros:fix-126-non_static_requestwrapper_cache?expand=1
- Remove the handler factory cache and just try to resolve the correct handler from the container on each send. (like notifications). https://github.com/jbogard/MediatR/compare/master...remcoros:fix-126-remove_factory_cache?expand=1

I would prefer option 2, let me know what you think.

jbogard:
So the reason I put the cache in is that I wanted to memoize which handler I needed to instantiate. But if the cache isn't even used lol

remcoros:
@jbogard If the handler is cached, it does mean that it's impossible to switch handlers trough the container during runtime.

For example:
Unbind(IRequestHandler<Ping>)
Bind(IAsyncRequestHandler<Ping>)
("Rebind" in ninject)

This is a highly unlikely scenario, but we do some funky stuff with Ninject and plugins where plugins can override each others implementation. Ie. plugin1 provides a handler, plugin2 overrides the handler.

Currently, we don't use mediatr handlers here. But we most likely will in the future. 

I'd prefer having the cache localized to the Mediator instance (instead of static). Or no cache at all (like how resolving notification handlers work).

[...]

PR #148, opened 2017-03-02
#136 - remove "initialized" variable

swimtver:
from "GetHandler" function & make private field in RequestHandlerImpl

### DIFF
--- a/src/MediatR/Internal/RequestHandler.cs
+++ b/src/MediatR/Internal/RequestHandler.cs
@@ -1,4 +1,4 @@
-using System.Collections.ObjectModel;
+﻿using System.Collections.ObjectModel;
 
 namespace MediatR.Internal
 {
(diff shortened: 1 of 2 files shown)


remcoros:
This still has the issue of the handler class instantiated twice on first use as discussed here: https://github.com/jbogard/MediatR/issues/136#issuecomment-279090119

I proposed to remove the caching all together and make it work like notifications (every 'Send' goes to the container). This is discussed here: https://github.com/jbogard/MediatR/issues/126

And a (proposed fix) here: https://github.com/jbogard/MediatR/compare/master...remcoros:fix-126-remove_factory_cache?expand=1

swimtver:
@remcoros, thanks for your answer. I believe that there is the reason, why author put handler instances and factory functions in cache. And "initialized" varailble's using looks like mistake.
 I looked at your fix, and i think it's a big change. It's gonna be design decision.

jbogard:
I introduced the cache so that you wouldn't have exceptions on every
resolution. This is all due to me making just one IRequest interface and
determining at runtime which handler actually matches. I can't determine
that except by trying to instantiate and caching the strategy on which
handler "works" (sync/async/async w/ cancellation token).

swimtver:
@jbogard, thanks for the detailed description. But you should look on LazyInitializer.EnsureInitialized using.
I think this code is doing not what you expected. Or I missed something?
I will try to write test to reproduce this behavior.

P14x:
The LazyInitializer.EnsureInitialized overload being used,
uses the boolean argument - which is always false - to determine whether the target has already been initialized.
Refactoring it into a field resolves this, so that the caching strategy works.

jbogard:
@remcoros I believe the caching would only fail for you if a user switched from one handler interface to another.

PR #174, opened 2017-05-30
Added the ability to send a request to all handlers and return all of their results.

Tsabo:
As my commit comment suggests, this addition is huge for me. We have multiple hardware platforms that we communicate with, from access control, alarm panels, cameras and recorders, elevators, etc... With this I can have a new "Doors" controller in my ASP.NET application with GetAll method. The controller doesn't need to know about the various door hardware providers, it just makes a request to MediatR for all of the doors and then my various Mercury, Assa, Axis, Isonas door request handler providers will return all of the doors that it has configured.

### DIFF
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -1,4 +1,4 @@
-using System.Threading;
+﻿using System.Threading;
 using System.Threading.Tasks;
 
 namespace MediatR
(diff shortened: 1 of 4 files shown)


jbogard:
How does this work with the pipeline? Does the pipeline wrap each handler or ALL handlers?

Tsabo:
It wraps each handler. I do not have a preference for this right now, so I'm more than happy to hear your advice on how it should be or how you would like it.

Here is an example of the output of a LINQPad test script.
image

bbrandt:
This is interesting.

This seems like a [Composed Message Processor](http://www.enterpriseintegrationpatterns.com/patterns/messaging/DistributionAggregate.html) or a simple [Recipient List](http://www.enterpriseintegrationpatterns.com/patterns/messaging/RecipientList.html)/[Aggregator](http://www.enterpriseintegrationpatterns.com/patterns/messaging/Aggregator.html).  Do you think these enterprise integration patterns would be appropriate in MediatR, which claims to be a "Simple, unambitious mediator implementation" or at another abstraction level?

It's been a few months since the PR was opened. Are you running a copy of this PR in production?  Are using the default `PublishCore` that calls `Task.WhenAll()`?  When you await `SendAll()` and multiple exceptions are thrown, only the first exception is raised from the await.  In your app code do you then swallow that exception and read all exceptions from the task returned from `SendAll()`?

src/MediatR/IMediator.cs
```diff
@@ -26,6 +26,15 @@ public interface IMediator
         Task Send(IRequest request, CancellationToken cancellationToken = default(CancellationToken));
 
         /// <summary>
+        /// Asynchronously send a request to a all request handlers and await for all of their replies
```
nikitadmitry: "a" article is excess.

jbogard:
This is better suited for a custom pipeline.

PR #177, opened 2017-06-13
Include build support for netstandard2.0

mrpmorris:
I simply added netstandard2.0 to the list of TargetFrameworks for the main project so I can use it on .NET Core 2


adamhathcock:
Since it targets `netstandard1.1` it should work.  .NET Core 2.0 implements `netstandard1.1` as well as all other standards right?  I haven't looked at the preview myself yet.

henkmollema:
The library should already work on .NET Standard 2.0, as 2.0 implements 1.1 as well. You would only need cross-compiling if you want to use 2.0 specific features.

jbogard:
Yes, that's the whole point of netstandard, is there something specific in netstandard2.0 we're trying to target? It didn't look like it.

PR #204, opened 2017-10-27
- adds contextual notification/request handlers

oguzhaneren:
- adds pipeline for notifications
- adds contextual notification/request handlers

You can pass data to handler like this:
```csharp
var context = new DefaultMediatorContext()
                                         {
                                             Items = new Dictionary<object, object>()
                                                     {
                                                         {"created-at",DateTime.UtcNow}
                                                     }
                                         };
await mediator.Publish(new PingedAsync(), context: context);
```
then you can access from handler:
```csharp
 public class PingedAsyncHandler : IContextualAsyncNotificationHandler<PingedAsync>
    {
        private readonly TextWriter _writer;

        public PingedAsyncHandler(TextWriter writer)
        {
            _writer = writer;
        }

        public Task Handle(PingedAsync notification,IMediatorContext context)
        {
            return _writer.WriteLineAsync($"--- Got pinged async. {context?.Items["created-at"]}");
        }
    }
```

### DIFF
--- a/src/MediatR/IMediatorContext.cs
+++ b/src/MediatR/IMediatorContext.cs
@@ -1,0 +1,21 @@
+using System.Collections.Generic;
+using System.Threading;
+
+namespace MediatR
+{
+    /// <summary>
+    /// Context
+    /// </summary>
+    public interface IMediatorContext
+    {
+        /// <summary>
+        /// Cancellation Token
+        /// </summary>
+        CancellationToken CancellationToken { get; set; }
+
+        /// <summary>
+        ///  Gets or sets a key/value collection that can be used to share data within the scope of this request/notification.
+        /// </summary>
+        IDictionary<object, object> Items { get; set; }
+    }
+}
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -29,37 +29,41 @@ namespace MediatR
             _multiInstanceFactory = multiInstanceFactory;
         }
 
-        public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, CancellationToken cancellationToken = default(CancellationToken))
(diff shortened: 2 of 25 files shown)


src/MediatR/IMediatorContext.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Cancellation Token
+        /// </summary>
+        CancellationToken CancellationToken { get; set; }
+
+        /// <summary>
+        ///  Gets or sets a key/value collection that can be used to share data within the scope of this request/notification.
+        /// </summary>
+        IDictionary<object, object> Items { get; set; }
```
bbrandt: Make sure whatever you use as a key in this dictionary has an implementation for Equals() and GetHashCode() and that it's not mutable or you're going to have a bad day.

https://www.knightlycode.com/2017/06/21/back-to-the-basics-custom-types-in-c-dictionary-key/
https://blog.markvincze.com/back-to-basics-dictionary-part-3-built-in-gethashcode/

src/MediatR/Mediator.cs
```diff
@@ -29,59 +31,75 @@ public Mediator(SingleInstanceFactory singleInstanceFactory, MultiInstanceFactor
...
+        }
+
+        private static IMediatorContext SetDefaultContext(CancellationToken cancellationToken, IMediatorContext context)
+        {
+            context = context ?? new DefaultMediatorContext();
+            if (context.CancellationToken == default(CancellationToken))
+            {
+                context.CancellationToken = cancellationToken;
+            }
+            return context;
```
bbrandt: I assume it's pretty minimal, but have we quantified the performance impact of this additional step in the pipeline when the set of notifications is empty?

jbogard:
Big merge conflict :)

PR #206, opened 2017-11-13
Updated IRequest inheritance heirarchy

JordyWells:


d-miles:
This would solve a problem I'm currently having in my own project! +1

jbogard:
What problem is this PR solving?

JordyWells:
It allows for the ability to implement generic type constraints on IRequest and IRequest<T>. It seems the inheritance is backwards when IRequest inherits from IRequest<unit>, Or maybe I just can't figure out why it was implemented in that manner. With the current structure generic type constraints aren't fully possible, I'm hoping this will fix that.

jbogard:
Can you add a test to show this?

JordyWells:
I have added some type constraints test, I am not a 100% sure if that is what you were after, they are more of a really basic example of what i am trying to do with type constraints.

JordyWells:
As an FYI here is the initial change where `IRequest` began inheriting `IRequest<Unit>`

https://github.com/jbogard/MediatR/commit/9a542f44490c0ab33a7bc6aec13b7fce75f409de

But this pull request does the same thing, but allows for the type constraints of `IRequest` 
unless I am missing something?

jbogard:
I’ve gotten some requests for changing the core interfaces so I want to
make sure it’s also captured via a test the core scenario these changes
enable.

JordyWells:
@jbogard I think the tests I have added should do that.

remcoros:
I ran into this as well when playing around with a generic domain event handler. 'fixed' it by using IRequest<unit> and some custom ninject registration/scanning.

I think I ran into issues with constraints as well. Some of it is here: https://github.com/remcoros/MediatRFun/blob/master/MediatR.DomainEvents/CommandHandler.cs

It makes sense that an IRequest<T> is an IRequest, like IEnumerable<T> is an IEnumerable, and the support for 'return void' handler is there.

jbogard:
With this change, is it now possible to call the wrong "Send"? And create the wrong "Handler"?

```c#
public class Foo : IRequest<Bar> { }

public class FooHandler : IRequestHandler<Foo> {
    public void Handle(Foo request) {
    }
}

var bar = await mediator.Send(foo);
// bar is null because I could create a request handler that returned nothing
```

Conversely I can call the wrong "Send" and not get the result back since `IRequest<TResult> : IRequest`.

Fundamentally the relationship seems backwards. Yes, `IEnumerable<T> : IEnumerable` but that's only because the values go from `object` to `T`. Not from `void` to `T`. I can't see `T : void` but rather `void : T` we just don't have a special `void` type in C# like there are in other languages.

JordyWells:
@jbogard ah, ok that makes sense. If i was to change it instead to use a common interface between the two, something like `IBaseRequest` that `IRequest<out TResponse>` inherits from. This would allow for type constraints of at least a shared type. Would that be a better solution?

It would then return all the original functionality, while still allowing me to specify a type constraint of `IBaseRequest` 

I will update the pull request to show an example of it.

test/MediatR.Tests/GenericTypeConstraintsTests.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            // Assert it is of type IRequest<T> but not IRequest
+            Assert.False(genericTypeConstraintsResponseReturn.IsIRequest);
+            Assert.True(genericTypeConstraintsResponseReturn.IsIRequestT);
+            Assert.True(genericTypeConstraintsResponseReturn.IsIBaseRequest);
+
+            // Verify it is of IRequest<Pong>
+            genericTypeConstraintsResponseReturn.Handle(ping)
+                .Select(x => x.ShouldBeOfType<IRequest<Pong>>())
+                .Select(x => x.ShouldBeOfType<IRequest>())
```
remcoros: this should not be IRequest right?

JordyWells: Sorry this test was actually broken, I've updated it to actually test the scenario correctly now.
But now ping should not be be of `IRequest`, it should be if of `IRequest<Pong>` and `IBaseRequest`

the updated version now accurately shows this.

PR #214, opened 2017-12-01
Added reference type constraints

JordyWells:


jbogard:
Would it be better to have it check for `default(TNotification)`? I'm a little wary of *adding* more generic constraints.

JordyWells:
I think that is 100% fair!

JordyWells:
@jbogard  Just thinking, would `default(TNotification)` be correct? It means you just cant send an empty version of an `INotification` implementation. but null would still be allowed. If we constrain it to class it means that the null check would still make sense, and that `INotification` must be of a reference type. So basically don't send ints or structs. Would it make sense to actually keep the class constraint, but actually add the `default(TNotification)` check as well, and throw a different exception?

JordyWells:
Just a second though, by checking `default(TNotification)` what if an implementations default value was actually something you did want to publish?
I think the class constraint ensuring its a reference type and not null might make the most sense?

jbogard:
What's the purpose for the class constraint? What scenario are we trying to prevent here? Sending value types? Is that in and of itself bad?

Won't this already fail if there aren't any handlers? Or not for something like `mediator.Publish(5)`?

JordyWells:
@jbogard You're actually correct, sending a value type like an int would already fail, but without the class constraint I can implement a struct and send that, which I don't think is wanted behavior?

jbogard:
You can send anything, as long as there is a handler. And I already check for that and throw an exception.

JordyWells:
Ok sounds good to me.

PR #219, opened 2017-12-14
Added 2  new Send method overloads

CarlosNetMty:
I think it could be helpful to have these new overloads that I added to Mediator.Send, which it allows users to be more specific with the request type that they are sending. 

Since I have been trying to get a project to be as segregated as possible I didn't want to have handlers on my UI project and I do not want to lose the ability to use my ViewModels as MediatR requests, I was thinking that maybe we could segregate this by using interfaces (ICreateUser : IRequest) but since MediatR only uses request.GetType() to understand the type of request this is not possible. These new overloads could make that scenario possible.

### DIFF
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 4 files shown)


jbogard:
I had this early on, but found it annoying to have to specify both the request and response type. Could you do:

```c#
var request = new Ping();
var response = mediator.Send((IPing) request);
```
?

Or, your handler can be made to be generic:

```c#
public class PingHandler<TRequest, TResponse> : IRequestHandler<TRequest, TResponse>
   where TRequest : IPing
{
    public Task<Pong> Handle(IPing request, CancellationToken cancellationToken)
    {
        return Task.FromResult(new Pong { Message = request.Message + " Pong" });
    }
}
```

This way you don't have an overload that is so explicit in the generic arguments.

jbogard:
Closing due to inactivity

PR #225, opened 2018-01-14
Updated LightInject to 5.1.2 adding support for ordered services

seesharper:
This PR updates LightInject to 5.1.2 adding support for ordered services.
This should complete the score card for for LightInject 👍

### DIFF
--- a/samples/MediatR.Examples.LightInject/MediatR.Examples.LightInject.csproj
+++ b/samples/MediatR.Examples.LightInject/MediatR.Examples.LightInject.csproj
@@ -13,7 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


dadhi:
Interesting, so the order of registrations themselves is not enough?

seesharper:
@dadhi Scanning an assembly is probably not going to bring back types in a deterministic way, so I figured an explicit API for dealing with the ordering was the best choice, at least for LightInject :)

dadhi:
OK, then just the clarity and discoverability.. nice.

jbogard:
It looks like this results in double publishing? https://ci.appveyor.com/project/jbogard/mediatr/build/93-eamslrgn#L237

seesharper:
@jbogard Should be good now 👍

PR #250, opened 2018-04-03
Consolidate to a single factory delegate

### LINKED ISSUE #247: Replace MultiInstanceFactory with collection injection
Given that "all" modern containers support `IEnumerable<TService>` injection, it may be better to use it instead of `MultiInstanceFactory`.

The main benefit is not the less code (though it is good), but removal of additional resolution roots, replacing them with collection dependency in scope of the same object-graph. So the less graph splits and service locators to serve them :)

The particular benefit will be for injecting collection of middlewares registered as open-generics. For example, DryIocZero then won't need this code https://github.com/jbogard/MediatR/blob/99b77d8ad43865bf320c5b85c1c7e1af108ba2e3/samples/MediatR.Examples.DryIocZero/Registrations.ttinclude#L43

jbogard:
And just having a single factory method.

Also consolidates some interfaces, removing the separate interface for void-based commands. Everything is `IRequest<T>` now, with just helper classes for void ones.


khellang:
Yes! 😍

src/MediatR/IRequestHandler.cs
```diff
@@ -29,32 +29,35 @@ public abstract class AsyncRequestHandler<TRequest, TResponse> : IRequestHandler
...
+            return Unit.Value;
+        }
 
         /// <summary>
         /// Override in a derived class for the handler logic
         /// </summary>
         /// <param name="request">Request</param>
         /// <returns>Response</returns>
-        protected abstract Task HandleCore(TRequest request);
+        protected abstract Task Handle(TRequest request);
```
khellang: Finally! :sparkles:

Y NO CANCELLATIONTOKEN THO?

jbogard:
You have an interface already! These are for no tokens.

khellang:
I thought those were just to avoid having to return `Unit`. Why are you throwing away tokens if you already have them?

jbogard:
Interfaces == async w/ cancellation token, `AsyncRequestHandler` == async w/o cancellation token, `RequestHandler` == sync (wrapping `Task.FromResult`). The abstract classes also ditch `Unit` but that's not their main purpose.

khellang:
Yeah, my beef is `AsyncRequestHandler` without cancellation token. Why would you ever disregard a cancellation token? I'm just curious.

jbogard:
I think originally these were here just to help with the breaking changes of adding the cancellation token. But if there's not real point I'll just get rid of them. It's a major release anyway.

PR #251, opened 2018-04-04
WIP: #247 Replace MultiInstanceFactory with collection injection

dadhi:
Tests passed; Samples work except for Unity and Ninject - throws exceptions for constraints support (may be not worth to fix them).

changed: RequestHandlerWrapperImpls are renamed to RequestProcessor and made public
changed: NotificationHandlerWrapperImpl is renamed to NotificationProcessor and made public
changed: RequestProcessor and NotificationProcessor have directly injected dependencies instead of service factories
changed: Activator.CreateInstance usages replaced with resolving RequestProcessor and NotificationProcessor via service factory, those are the only things service factory is used for (may be simplified further?)

removed: MultiInstanceFactory
removed: Mediator._voidRequestHandlers, _requestHandlers, _notificationHandlers cache - relying on IoC cache/lifetimes instead


dadhi:
@jbogard Hello, 

That what I meant about replacing Resolution with Injection, not just the replacing resolution of
 many with resolution of `IEnumerable<T>` as in your branch.

I have further simplification, but it requires converging of `IRequestHandler` intefaces. Here is the [branched commit](https://github.com/dadhi/MediatR/commit/e5af00748f859232bb5fbece450ef04b2596397d).

jbogard:
Ahhhh doesn't this make registration more complicated? It looks like most containers have to add the processors?

jbogard:
And it's still using resolution, just a middle type of the processors to help the containers, right?

dadhi:
Yes, but it uses only on a highest level. So you even no need to register service factory, just pass to the Mediator.

Also SM and Lamar don't need to be changed. DryIoc need to be changed only because ..Processors now are registered as concrete types. If I add interfaces `IRequestProcessor` and `INotoficationProcessor` then convention will kick in.
Also the containers are already registering bunch of MediatR infrastructure.

jbogard:
Don't all the containers still have to register the service factory? I guess comparing I'm seeing more registrations, not fewer, for each container.

dadhi:
Let me check, may be I can remove unnecessary bits.

But IMHO having two explicit integration roots/points is a good thing. 

Moreover, from DI standpoint, it is better to have single inspectable dependency tree, instead of one cut in pieces by service locators.

May be it can be made more user API friendly, let's see.

JohanKlijn:
@dadhi : if you remove the MultiInstanceFactory there will be probably an issue when you use Unity as the IoC container. Because it that case it does not support covariant notification handlers. See one of the last commits in the Unity example. (https://github.com/dadhi/MediatR/commit/b15051d881d480ea2732e51d2b8d8b4d8fe5fb8b)

dadhi:
@JohanKlijn 
Yes, I know. May be it woulf be possible to introduce some kind of fallback. 
But hopefully Unity will support this feature soon. Do you know is it planned?

Anyway, it is an experiment how far I can go. 
MediatR is good compact project fpr that, to find a better ways of integration with modern DI libs.

jbogard:
I'm totally stealing that `IRequestHandler` consolidation. Many containers mess up because you've got two kinds of interfaces there.

dadhi:
Will provide a new one based on the 5 alpha.

JohanKlijn:
@dadhi : FYI, I created a issue at the Unity repo related to contravariant support, see https://github.com/unitycontainer/unity/issues/215. Maybe we get some information when/if they are going to support this.

PR #253, opened 2018-04-09
Consolidating some handlers; renamed some things for posterity

jbogard:
Just for you @khellang

### DIFF
--- a/src/MediatR/INotificationHandler.cs
+++ b/src/MediatR/INotificationHandler.cs
@@ -25,9 +25,9 @@ namespace MediatR
     public abstract class NotificationHandler<TNotification> : INotificationHandler<TNotification>
         where TNotification : INotification
     {
-        public Task Handle(TNotification notification, CancellationToken cancellationToken)
+        Task INotificationHandler<TNotification>.Handle(TNotification notification, CancellationToken cancellationToken)
         {
-            HandleCore(notification);
+            Handle(notification);
             return Unit.Task;
         }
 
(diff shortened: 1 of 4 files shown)


src/MediatR/INotificationHandler.cs
```diff
@@ -27,32 +27,14 @@ public abstract class NotificationHandler<TNotification> : INotificationHandler<
...
-        public Task Handle(TNotification notification, CancellationToken cancellationToken)
-            => Handle(notification);
-
-        /// <summary>
-        /// Override in a derived class for the handler logic 
-        /// </summary>
-        /// <param name="notification">Notification</param>
-        /// <returns>A task</returns>
-        protected abstract Task Handle(TNotification notification);
+        protected abstract void HandleCore(TNotification notification);
```
khellang: Why did these become `HandleCore` again?

khellang: Also, the async notification handler should probably also forward the cancellation token?

jbogard: So the `AsyncRequestHandler` class now has an overload that differs only in return type. I could:

 - Make all the Handle methods be explicit interface implementations, so I can keep the overload
 - Have only this class have `HandleCore`
 - Get rid of all these helper classes because does it really matter

khellang: I'd go for

> Make all the Handle methods be explicit interface implementations, so I can keep the overload

😄

PR #277, opened 2018-06-06
change container.Register to container.Collection.Register

ojraqueno:
Updated for the Pipeline types. This fixes the SimpleInjector error:

The supplied list of types contains one or multiple open generic types, but this method is unable to handle open generic types because it can only map closed generic service types to a single implementation. You must register the open-generic types separately using the Register(Type, Type) overload. Alternatively, try using Container.Collection.Register instead, if you expect to have multiple implementations per closed-generic abstraction.


jbogard:
Ohhhhhh ok, that makes sense. I had a bunch of "obsolete" errors come up but it didn't tell me how to fix it so I just guessed. I guessed wrong lol

ojraqueno:
Thank you, Jimmy :) This is my first pull request so this is a milestone for me!

PR #293, opened 2018-07-13
Configurable task execution

bartosz6:
Proof of concept
- tests are passing
- sample projects not yet updated - failing


bartosz6:
the simpler, the better https://github.com/jbogard/MediatR/pull/292/files

Issue #299, opened 2018-07-30
Weakly typed Publish variant


robbaman:
Sorry if this has been asked before, but is there a way to publish an event using a `Type` object instead of a strongly typed call?

I'm storing events in a database and in event of a failure these events are sent again after some time. At that point I only have a `Type` object and a deserialized object (of that particular type). I want to publish this event using MediatR, but the `Publish<T>(T notification)` method does not seem to have an overload that accepts a `Type` and an object of said type.

Am I missing something or is this just something that I'm going to have to write something custom for? Didn't see any mention of this in the Wiki and there doesn't seem to be an overload for this.

jbogard:
There's not one for Send, either. This does seem useful - pull request perhaps?

bartosz6:
Well I can't think of any other solution than creating some wrapper for your events, but it would require some custom handlig. Would be easier to overload Publish.

Issue #124, opened 2017-01-05
Publishing collection of INotifications


jblackburn21:
In v2.1 I was able to loop through a collection of INotifications and publish the events since it used `var notificationType = notification.GetType();` to determine the exact notification type of the handlers to execute.  In v3, it uses `var notificationType = typeof(TNotification);`, which causes it to look for handlers that implement IAsyncNotificationHandler<INotification>.  Can this be changed back to the original behavior?

jbogard:
Not sure what the problem is, do you have a snippet that shows the bad behavior? I want to write a test for this and I'm not sure what I'm aiming for.

jblackburn21:
Here is a variation of the unit test.  I collect a list of events and then publish them at the end of my process, so the list could contain different types.

```
[Fact]
        public async Task Should_resolve_handler_for_list_of_generic_notifications()
        {
            var builder = new StringBuilder();
            var writer = new StringWriter(builder);

            var container = new Container(cfg =>
            {
                cfg.Scan(scanner =>
                {
                    scanner.AssemblyContainingType(typeof(AsyncPublishTests));
                    scanner.IncludeNamespaceContainingType<Ping>();
                    scanner.WithDefaultConventions();
                    scanner.AddAllTypesOf(typeof(IAsyncNotificationHandler<>));
                });
                cfg.For<TextWriter>().Use(writer);
                cfg.For<IMediator>().Use<Mediator>();
                cfg.For<SingleInstanceFactory>().Use<SingleInstanceFactory>(ctx => t => ctx.GetInstance(t));
                cfg.For<MultiInstanceFactory>().Use<MultiInstanceFactory>(ctx => t => ctx.GetAllInstances(t));
            });

            var mediator = container.GetInstance<IMediator>();

            var pings = new List<INotification>()
            {
                new Ping {Message = "Ping"}
            };

            foreach (var ping in pings)
            {
                await mediator.Publish(ping);
            }

            var result = builder.ToString().Split(new[] { Environment.NewLine }, StringSplitOptions.None);
            result.ShouldContain("Ping Pong");
        }
```

jbogard:
Ah, makes sense.

jblackburn21:
I was able to get it to work by adding the handler wrappers back in.  Let me know what you think.

https://github.com/jbogard/MediatR/compare/master...jblackburn21:feature/NotificationHandlerWrapper?expand=1

Also, thanks for fixing my other issue.  I was able to verify it after I got this working.

[...]

PR #300, opened 2018-07-31
Weakly typed publish overload

bartosz6:
Created an overload for publish method. #299

### DIFF
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -24,6 +24,15 @@ namespace MediatR
         /// <param name="notification">Notification object</param>
         /// <param name="cancellationToken">Optional cancellation token</param>
         /// <returns>A task that represents the publish operation.</returns>
-        Task Publish(INotification notification, CancellationToken cancellationToken = default);
+        Task Publish(object notification, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Asynchronously send a notification to multiple handlers
+        /// </summary>
+        /// <param name="notification">Notification object</param>
+        /// <param name="cancellationToken">Optional cancellation token</param>
+        /// <returns>A task that represents the publish operation.</returns>
+        Task Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
+            where TNotification : INotification;
     }
 }
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -24,15 +24,6 @@ namespace MediatR
         /// <param name="notification">Notification object</param>
         /// <param name="cancellationToken">Optional cancellation token</param>
         /// <returns>A task that represents the publish operation.</returns>
-        Task Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
-            where TNotification : INotification;
-        /// <summary>
-        /// Asynchronously send a notification to multiple handlers
-        /// </summary>
-        /// <param name="notification">Notification object</param>
-        /// <param name="notificationType">Type of notification object</param>
-        /// <param name="cancellationToken">Optional cancellation token</param>
-        /// <returns>A task that represents the publish operation.</returns>
-        Task Publish(INotification notification, Type notificationType, CancellationToken cancellationToken = default);
+        Task Publish(INotification notification, CancellationToken cancellationToken = default);
     }
 }
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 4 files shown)


src/MediatR/IMediator.cs
```diff
@@ -25,5 +26,13 @@ public interface IMediator
...
         Task Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
             where TNotification : INotification;
+        /// <summary>
+        /// Asynchronously send a notification to multiple handlers
+        /// </summary>
+        /// <param name="notification">Notification object</param>
+        /// <param name="notificationType">Type of notification object</param>
+        /// <param name="cancellationToken">Optional cancellation token</param>
+        /// <returns>A task that represents the publish operation.</returns>
+        Task Publish(INotification notification, Type notificationType, CancellationToken cancellationToken = default);
```
jbogard: Can't you infer the notification type from the notification itself? And I would assume that the weakly-typed would mean `object`, not `INotification`.

If you look at the `Send` method, it infers the request type from the request object passed in. I would expect this method to look like:

```
Task Publish(object notification, CancellationToken cancellationToken = default);
```

That way it's very obvious what you're trying to do here. I think callers wouldn't want to have to call `GetType` before publishing.

src/MediatR/Mediator.cs
```diff
@@ -41,14 +41,19 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
         }
 
         public Task Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
-            where TNotification : INotification
+            where TNotification : INotification => Publish(notification, notification?.GetType(), cancellationToken);
```
jbogard: This could be `typeof(TNotification)`

src/MediatR/IMediator.cs
```diff
@@ -23,7 +24,6 @@ public interface IMediator
         /// <param name="notification">Notification object</param>
         /// <param name="cancellationToken">Optional cancellation token</param>
         /// <returns>A task that represents the publish operation.</returns>
-        Task Publish<TNotification>(TNotification notification, CancellationToken cancellationToken = default)
-            where TNotification : INotification;
+        Task Publish(INotification notification, CancellationToken cancellationToken = default);
```
jbogard: This might still require a cast on the caller side - we can mitigate this by checking to see if the object is an `INotification`, and throw an `ArgumentException` otherwise.

The overload rules are weird too - the generic method might never be called because the compiler assumes this overload.

bartosz6: Please check: https://github.com/jbogard/MediatR/pull/300#discussion_r206524326

About overload - it is redundant with generic, so generic one has been removed. It should be brought back if we decide to go for `object` instead of `INotification`

src/MediatR/Mediator.cs
remcoros: should be GetType(), as TNotification can be a interface/abstract type.

src/MediatR/Mediator.cs
```diff
@@ -41,20 +41,34 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
+             where TNotification : INotification
         {
             if (notification == null)
             {
                 throw new ArgumentNullException(nameof(notification));
             }
-
-            var notificationType = notification.GetType();
+            
+            var notificationType = typeof(TNotification);
```
remcoros: this should really be GetType() like it was, as TNotification can be an interface/abstract type. Especially in this case, since in the added Publish method, the notification is checked and casted to INotification, so TNotification == INotification and it wouldn't find any handlers.

Note this isn't covered by any tests, but has been brought up before.

jbogard: I don't think the generic version can use the overload. The generic version needs to use `typeof(TInterface)` and the non-generic one needs `notification.GetType()`

It's a separate issue to default to GetType - I'd rather not.

remcoros: See https://github.com/jbogard/MediatR/issues/124

It needs the concrete type, not the type of T decided by the compiler.

e.g.

```cs
var events = new IEvent[] { new EventA(), new EventB(), new EventC() };
events.Select(e => _mediatr.Publish(e));
```

TNotification is IEvent there, that's not going to work.

remcoros:
Looking at the PR as it is now:

    Task Publish(object notification, CancellationToken cancellationToken = default);

this does not take a 'Type'. What is the benefit here? Why not call the generic version.

```cs
object obj = Deserialize(somestream);

_mediatr.Publish(obj);
```

as long as the underlying type of 'obj' is known, this already works.

remcoros:
I found it weird this passes the unit tests, so did some digging @jbogard 

https://github.com/jbogard/MediatR/commit/0412db8ef1e893a350fb3a8788527fe89dacfa46#diff-dd702b31bb5bd69f10b2c5ce72015121R99

this broke the essence of the test by removing the explicit declaration of the event as INotification. I should have put a comment there :-/

I can add a better test for it.

anyway, using typeof() would be a regression.

src/MediatR/Mediator.cs
```diff
@@ -41,18 +41,28 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
         {
             if (notification == null)
             {
                 throw new ArgumentNullException(nameof(notification));
             }
 
-            var notificationType = notification.GetType();
-            var handler = _notificationHandlers.GetOrAdd(notificationType,
-                t => (NotificationHandlerWrapper)Activator.CreateInstance(typeof(NotificationHandlerWrapperImpl<>).MakeGenericType(notificationType)));
+            return Publish(notification, typeof(TNotification));
```
jbogard: My bad. Sorry last request - as @remcoros pointed out, using `typeof(TNotification)` would be a regression. Sorrryyyyyy

jbogard: So we'd do `notification.GetType()` here instead. Sorry again!

bartosz6: I'll include some tests for this scenario :)

PR #320, opened 2018-10-17
Overwriteablehandler: OverwriteableAttribute

Condra963:
Hi,

I Created an opportunity to overwrite handlers. You are probably thinking, in what scenario will you ever do this?

Well let me explain.

I have **packages** that are developed that i want to use all over my **software**. There i use `MediatR` to call my `EFCore` handling. Now i have scenario's where i do not want to keep all of my entities, some i want to expand with more details because that's related only to that software. But my basic package will still do the same calls that i did before but these calls should not handle Entity, they should handle `ExtendedEntity`. So i needed a **new** handler.

Now for this scenario, i created handlers based on the `interface` of my `Entity` and not based on the `Entity` himself, else what i am telling before will not work. So i have for example an `IRequest<IPong>` named `OverwriteablePingRequest`. 
In the normal way without changing the current `DbContext` with extended entities it will just call `OverwriteablePingRequestHandler`.
When i extend the entity `Pong` with `ExtendPong` i also become and `ExtendedDbContext` so i want to call `OverwrittenPingRequestHandler` with the same `OverwriteablePingRequest`.
With MediatR you can load multiple assemblies where it will register all your handlers with your requests, but what comes first will be used first. 
So you could just replace every service in your `ServiceCollection` with the handlers that should be overwritten, true, but what if you have more then 1000Handlers that should be overwritten? So why are we here? To make this happen automatically.

In this pull request you only see the `Attribute` that i created to make you say which `Handler` can be overwritten. If you add this `OverwriteableAttribute` and it finds a second `Handler` that suits your `Request` it will overwrite the `Handler` with the `OverwriteableAttribute`.

**I need this pull-request to be accepted if you want to build and test run the changes in MediatR.DI.**

I also have a couple of questions:
- Does the naming suits you? If not could you just modify it to a proper naming that suits you instead of denying my pull-request please? Because i could use this as quickly as possible.
- Do you approve to let it work this way with the attribute or do you have other suggestions?

I you have questions for me, just ask.

Thanks.

### DIFF
--- a/src/MediatR/Attributes/OverwriteableAttribute.cs
+++ b/src/MediatR/Attributes/OverwriteableAttribute.cs
@@ -1,0 +1,8 @@
+using System;
+
+namespace MediatR.Attributes
+{
+    public class OverwriteableAttribute : Attribute
+    {
+    }
+}


lilasquared:
@Condra963 Why does this belong in MediatR? You don't have anything in MediatR using this Attribute.  It seems this attribute is more suited to exist in your own project.

Condra963:
@lilasquared can you please read the complete explenation? Because i need it in MediatR.DI. But an Attribute should not be defined in MediatR.DI, it does not belong there.

lilasquared:
@Condra963 why wouldn't the attribute just exist in that project then?

lilasquared:
what is your reason for "it should not be defined in MediatR.DI"?

Condra963:
Hi

It is because an Attribute does not require other dependencies, so it should be in the core project.
Also because if you want to create a base handler like `OverwriteableRequestHandler` like you have `AsyncRequestHandler` it is also required in this assembly to prevent loop references.
And it is also needed when you create an alternative for MediatR.DI where you also want to use the same attribute.
Actually if i understand it correctly, this attribute should be here by design.

I hope my explanation is what you need to approve this?

lilasquared:
I can't approve this at all I'm just asking questions.  Are you particularly blocked in your solution because this is not a feature of MediatR or have you solved it for your use case?

lilasquared:
Yes what it does is very clear, but why it should be part of this project and the DI project is not clear to me.  It appears to me to be a very specific architecture that your software has.  I am not sure if the project benefits much from having this overriding logic.  Just my two cents though as I am not the project maintainer.

Condra963:
Let me explain again in another situation.
You use a package of nuget in your software, that one is using MediatR en does some handling. But you want to perform some other actions when one of these handlers is called, you can overwrite this handler and add extra stuff to it.
I really do not understand why you are being so difficult about this, being able to write a handler that is called from code that you are not able to modify yourself is useful in multiple scenarios.

jbogard:
Why can't you use pipeline behaviors?

Condra963:
How would pipeline behavior work to overwrite an existing handler that is already been written?

lilasquared:
Pipeline behaviors fire before the handler, so you can skip the handler processing altogether.  The issue of registering thousands of them would still be a problem, especially registering them at the right point so they don't skip any other pipeline behaviors that may be in place.

jbogard:
Because you decide whether or not to call "next" in a pipeline behavior:

```c#
public class OverridingBehavior<TRequest, TResponse> : IPipelineBehavior<TRequest, TResponse> {
    public async Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<TResponse> next) {
        await Logger.InfoAsync($"Handling {typeof(TRequest).Name}");
        TResponse response;
        if (MyApp.Reasons)
            { response = await next(); }
        await Logger.InfoAsync($"Handled {typeof(TResponse).Name}");
        return response;
    }
}
```

Condra963:
But the handler already exists in the package that i depend on and it will be registered no matter what i do, these handlers will be registered by default. So if i call my request it will call a pipeline and that will just call the handler that is defined by the package and then i can choose if i want to call mine as well if i understand correctly? But i want to replace the handler of the package, i only want my handler to be called. What about that?

jbogard:
You can also replace their handler in the DI configuration. The `ServiceCollection` is mutable, so you can find their registered instances, remove them, then add your own.

lilasquared:
@Condra963 instead of having a `OverwrittenHandler`, the work for that would be in the pipeline behavior in your project.  
You would skip calling `next()` at all.  `next()` is where the original handler would be called.

Condra963:
Yes i know, and that is what i tell in my current explanation that i know that you can do this, but with my solution i just need to define my assembly and it will be defined correctly without all that extra work. I do not need pipelinebehavior what in my opinion is to do call after call after call after call instead of overwriting something. And it's also extra unnecessary work because like i say my solution just pick the handler you want and register in into the dependencyinjection container.

And than i'm still wondering, are you sure with pipelinebehavior will catch my scenario? Because the request i call is one and the same, it only goes to another handler. You will still have the same problem and that is i cannot choose which handler i want to register to my container that the current request should call.

Condra963:
And i am looking for a solution that if i just derive from that current handler that it will take the handler that derives from the current handler. In that scenario you do not need the attribute, but the Handle method should be a virtual. I don't know what you prefer?

jbogard:
Why can't you override the container configuration? What container are you using?

lilasquared:
@jbogard the actual change is in https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection/pull/46

jbogard:
Ah OK. But `ServiceCollection` already lets you do `Replace`, so you can *already* replace handlers.

Condra963:
I do not say that i cannot override the container, i say it is not necessary with my solution. In a scenario where you want to do this for 1000Handlers i only have 1 line code and else i will have 1000. So my choice was made fast. I really do not understand why you are being so difficult about this. I take care of the fact that you shouldn't replace every service in you container. It looks before registering which one it wants to use. Instead of people everytime in every project that they want to overwrite a handler need to do servicecollection.ReplaceService (x1000 times) it is already taken care of in MediatR. Why shouldn't this be useful at all? And if you 1000 project where you have 1000handlers that you want to overwrite you have 1 000 000 handlers to overwrite and 1 000 000 lines to write that can be easily just take care of in MediatR.DI. I really do not get it that you people do not see the use of it.

jbogard:
Well, I've never needed to do this, and since I write my OSS for my projects/clients first, I have quite a high bar for features I'd never use, *especially* if adding a feature only to support a particular way of registering.

There could be a simpler solution here. Right now when you pass in assemblies, the DI project doesn't do any checking to see if there are already handlers registered. One solution would be "if there's already a handler defined, overwrite it". We now have a config object, so it's be pretty easy to turn this kind of behavior off/on.

jbogard:
Also, I assume that whether there are 1 or 1000000 handlers to override, you would have the same amount of code to apply your override rules. Just like the DI package scans assemblies for types you could do the same.

remcoros:
Having an attribute in MediatR core project without any code using it, is unnecessary. It would be logical if MediatR was split up in two projects, MediatR/MediatR.Abstractions. But even then, it will be unused code.

I think the better solution would be to create a separate packages for this, like 'MediatR.OverwritableHandler.DI / MediatR.OverwritableHandler'. Where in one the attribute will live, in the other the code to link it to a DI container.

Condra963:
**Package that i use**
- `OverwritableHandler1 : IRequestHandler<Request1, IResult1>`
- `OverwritableHandler2 : IRequestHandler<Request2, IResult2>`
- `OverwritableHandler3 : IRequestHandler<Request3, IResult3>`

**My software**
- `OverwrittenHandler1 : IRequestHandler<Request1, IResult1>`
- `OverwrittenHandler2 : IRequestHandler<Request2, IResult2>`
- `OverwrittenHandler3 : IRequestHandler<Request3, IResult3>`

When two handlers are registered with the same ServiceType (interface) and `Request1` is called, the handler that is resolved (the base one, or the overwritten one) is undefined. What I want is that, by design, the most derived handler is called (or only the most derived is registered in the first place).  
So for example in my software, At the moment I have to replace all handlers manually like this:
- `serviceCollection.Replace(new ServiceDescriptor(typeof(IRequestHandler<Request1, IResult1>), typeof(OverwrittenHandler1), ServiceLifeTime.Transient)`
- `serviceCollection.Replace(new ServiceDescriptor(typeof(IRequestHandler<Request2, IResult2>), typeof(OverwrittenHandler2), ServiceLifeTime.Transient) `
- `serviceCollection.Replace(new ServiceDescriptor(typeof(IRequestHandler<Request3, IResult3>), typeof(OverwrittenHandler3), ServiceLifeTime.Transient)`

My suggested change does this automatically, saving complexity, time, and lines of code.

Maybe the attribute is not necessary and it could be handled like this:
**My software**
- `OverwrittenHandler1 : OverwritableHandler1`
- `OverwrittenHandler2 : OverwritableHandler1`
- `OverwrittenHandler3 : OverwritableHandler1`

When a handler that is derived from another handler is being registered, the base handler should be replaced (or not be added in the first place). So in the end only the most derived handlers are registered.


<br/>


I believe this is a useful feature, however, I am not sure if using an attribute is the best way to do it. I am also unsure if this is only a useful feature in Mediatr or if it could also benefit ServiceCollection, since the registering of multiple same ServiceTypes (interfaces) and the resolving of them is the responsibility of ServiceCollection.

jbogard:
Your way of registering could be *greatly* simplified.

Condra963:
Could you tell me how you simplify this? Because everything is registered through MediatR except the Handlers that wants to overwrite the basehandlers.

jbogard:
Use reflection just like I do in the DI package to scan for handlers in your assembly and register them.

Is the problem that the DI package doesn't allow you to override handlers? If so it's a small change to that library to allow that to be configurable in some way besides an attribute.

Condra963:
Yes, I know you can use reflection, but why should I write reflection for something that already exists and can be simply modified once? Now I need to copy this to every project over and over again or build some package between while the solution I suggest is a solution for everyone without any extra work.


The only thing I suggest is adding a non-breaking and simple change that takes care of registering the deriving handlers with the exact same interfacetype only once so that I don't have to deal with it in every project in the future. Because at this moment it's like a bug that you can register 2 times the same IRequestHandler<Request1, IRespond1> for example. Isn't?

I can throw the attribute away if you want like I said before. Then I will make a fix in MediatR.DI that looks for deriving handlers. But I need to know if you are considering to accept this pull request if I modify it, if not I will look for another solution for this.

Thanks.

jbogard:
No, I don't think the fix is in this library, it's how the DI package registers handlers. It's not a bug either - you're registering a handler twice. It's unsupported behavior.

I do agree that the behavior to allow some choice in handlers would work, but that's for the DI project. Can you create an issue, something like "Allow user to select/configure which handler to use for duplicate handlers"?

Condra963:
Ok, then i remove this attribute from this package and continue my work in MediatR.DI but big change i already made is in MediatR.DI, here is just the attribute and i did my explenation here about both pull-requests because else you would be thinking "why does someone add some attribute in here?".

But let us move the discussion to MediatR.DI where i will link my original pull-request in MediatR.DI of how we could implement this? And if we figure out what the best way to implement this is, i can modify my pull-request.

Just a little explanation why i see this more as a bug than unsupported behavior:
Registering 2 times the same interface in your servicecollection will shoot out the first it finds, so when registering 2 times the same IRequestHandler<Request1, IRespond1> servicecollection should already throw an exception because it makes the injection unreliable. It's like you have identical twins, they look the same but they don't answer the same if you ask them something so if it's called by DependencyInjection you have 50% change the right twin will answer your question with the answer you expect. But this would be more something that should be discussed in ServiceCollection behavior.

But whatever, discussion will be moved to: jbogard/MediatR.Extensions.Microsoft.DependencyInjection#47

PR #326, opened 2018-11-12
Add PipelineExtensions for IRequest and INotification

FuncLun:
PipelineExtensions.cs allows a pipeline style of programming rather than nested style. (This is not related to PipelineBehaviors.) This style is possible with F#'s Pipe Forward operator. And, there is discussion in adding Pipe Forward operators to C# in [csharplang](https://github.com/dotnet/csharplang/issues?q=74+96).

Instead of nesting:
```
Mediator.Send(
    new Request()
    {
        Value = "value",
    },
    token);
```
PipelineExtension:
```
new Request()
    {
        Value = "value",
    }
    .Send(Mediator, token);
```
Pipe Forward (proposed CSharp feature):
```
new Request()
    {
        Value = "value",
    }
    |> Mediator(@, token);
```
The discussions have suggested using `@` or `<|` (what F# uses).

Unit Tests included.

### DIFF
--- a/test/MediatR.Tests/PublishTests.cs
+++ b/test/MediatR.Tests/PublishTests.cs
@@ -182,5 +182,52 @@ namespace MediatR.Tests
             result.ShouldContain("Ping Pong");
             result.ShouldContain("Ping Pung");
         }
+
+        [Fact]
+        public async Task Should_resolve_handlers_given_interface_with_mediator_extnesion()
+        {
+            var builder = new StringBuilder();
+            var writer = new StringWriter(builder);
+
+            var container = new Container(cfg =>
+            {
+                cfg.Scan(scanner =>
+                {
+                    scanner.AssemblyContainingType(typeof(PublishTests));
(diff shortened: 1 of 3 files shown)


test/MediatR.Tests/PublishTests.cs
```diff
@@ -182,5 +182,52 @@ public async Task Should_resolve_handlers_given_interface()
             result.ShouldContain("Ping Pong");
             result.ShouldContain("Ping Pung");
         }
+
+        [Fact]
+        public async Task Should_resolve_handlers_given_interface_with_mediator_extnesion()
```
lilasquared:
```suggestion
        public async Task Should_resolve_handlers_given_interface_with_mediator_extension()
```

jbogard:
Very interesting! Does this have any use if C# doesn't have the pipe forwarding operator? Or would this be included to work better in F#?

I don't like adding speculative features is why I ask.

jbogard:
Closing for now, until C# gets a pipe forward operator.

PR #332, opened 2018-11-15
update NotificationHandlerWrapper to await handlers

lilasquared:
alternative to https://github.com/jbogard/MediatR/pull/330

fixes https://github.com/jbogard/MediatR/issues/328

### DIFF
--- a/src/MediatR/Internal/NotificationHandlerWrapper.cs
+++ b/src/MediatR/Internal/NotificationHandlerWrapper.cs
@@ -18,7 +18,7 @@ namespace MediatR.Internal
         {
             var handlers = serviceFactory
                 .GetInstances<INotificationHandler<TNotification>>()
-                .Select(x => x.Handle((TNotification)notification, cancellationToken));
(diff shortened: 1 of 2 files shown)


src/MediatR/Internal/NotificationHandlerWrapper.cs
```diff
@@ -18,7 +18,7 @@ public override Task Handle(INotification notification, CancellationToken cancel
         {
             var handlers = serviceFactory
                 .GetInstances<INotificationHandler<TNotification>>()
-                .Select(x => x.Handle((TNotification)notification, cancellationToken));
+                .Select(async x => await x.Handle((TNotification)notification, cancellationToken));
```
FuncLun: Should this await have `.ConfigureAwait(false)`?

lilasquared: if it is in PublishCore doesn't that work?  If it is here that can't be overridden (not sure if that matters)

PR #362, opened 2019-02-02
Added sample project to show the ability of wrapping MediatR to help remove direct dependencies on IRequest etc

JonyTek:
WHAT: A sample project showing off the ability to wrap MediatR to allow you to remove direct dependencies. 

WHY: I have Domain libraries which contain Commands and I would prefer to keep these libraries pure without external dependencies. Essentially I would not like to have references to IRequest, INotification etc within my Domain but prefer to be able to define my own interfaces.

SOLUTION: Define custom interfaces in my base Domain library along with things like Aggregate<>, ValueObject etc and use these to mark my Commands/ Queries and have things wire up based off these. 

I do think that this would be quite a common use case so think it would be beneficial to have some example code available to help people get up and running more quickly.

Thanks.

### DIFF
--- a/MediatR.sln
+++ b/MediatR.sln
@@ -45,7 +45,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MediatR.Examples.DryIocZero
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MediatR.Examples.Lamar", "samples\MediatR.Examples.Lamar\MediatR.Examples.Lamar.csproj", "{F014598D-8B85-4D7E-942A-3493107ABE43}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MediatR.Examples.PublishStrategies", "samples\MediatR.Examples.PublishStrategies\MediatR.Examples.PublishStrategies.csproj", "{867EBA13-62F4-4525-8F92-B0AD828EE6D4}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "MediatR.Examples.PublishStrategies", "samples\MediatR.Examples.PublishStrategies\MediatR.Examples.PublishStrategies.csproj", "{867EBA13-62F4-4525-8F92-B0AD828EE6D4}"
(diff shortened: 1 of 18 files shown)


jbogard:
Sharing domain assemblies as a nuget package *also* seems like a bad idea. Sharing domain events across service boundaries is also a bad idea.

Domain events are things internal to a bounded context, and deployed components inside a bounded context shouldn't need to use package managers for shared behaviors.

Domain events should not be shared between bounded contexts either - you need to separate the messages that are for internal use only and for external use - and those aren't domain events. Those are service-level events.

JonyTek:
So when I refer to a service I am talking about an independent component and not a business capability. For example I have a commands api who receives commands and then performs operations within the domain e.g. CreateContract. These domain events (ContractCreated) are then persisted into EventStore. I then have additional services who subscribe and maintain readmodels - MongoSynchroniser and ElasticSearchSynchroniser are examples of this. These synchronisers depend on these events within the Domain library. Each service is maintained within their own solution hence the reason to share the domain library.

I think we are confusing what I call a service with what you tend to refer to as components?

These events are not shared outside of a bounded context but are across components

jbogard:
Perhaps? I'm referring to services in the SOA/microservice sense. If your services are within a bounded context, then just reference projects directly. Don't go through NuGet, that adds needless overhead.

jbogard:
Just one other nitpick - microservices that share a domain model or database aren't microservices. Microservices are a service-oriented architecture with a bent towards the smallest autonomous boundary.

If these things you're deploying manage their own state, then yeah I'd call them microservices.

TheFellow:
FWIW @JonyTek wrapping the MediatR implementation was the first thing I did when I picked it up.

In my case I'm designing the contexts/components to be decoupled by messaging like MediatR provides, but i want to leave some space available in the event I do need to actually break apart a context into its own autonomous service. If that were the case then I would need to have a more robust messaging system like a Service Bus, but fundamentally the messages still need to arrive at the handler. It's currently unclear if that will be necessary at all, a permanent requirement, or a shorter lived stop-gap solution.

By wrapping MediatR I'm free to swap implementations and leave my core free to target the messaging bus port, and plug it with adapters as needed to exchange behavior.

PR #363, opened 2019-02-06
A slightly different Mediator.Send interface

guido:


guido:
This pull request is just a simple idea about simplifying Mediator Send code so that refelction is trimmed down. Sadly it requires a slightly different interface and I am quite sure that this trick is well known. Still, I think that t might be ok with the help of some syntactic sugar

jbogard:
This was one of my early attempts, but it's super annoying to have to specify the source/destination type. Hopefully C#'s generic inference will improve.

PR #368, opened 2019-02-17
Add CancellationToken to IRequestPostProcessor

dariogriffo:
I'm the need of using IRequestPostProcessor, and wanted to make sure I can pass all the way down the CancellationToken.
This most likely be a broken change?


jbogard:
Yep, breaking change. Can you bump the version with this too?

dariogriffo:
> Yep, breaking change. Can you bump the version with this too?

Done

PR #401, opened 2019-05-06
Small refactoring

ralmsdeveloper:
Here is my small contribution to this wonderful project.

It seems to be a very simple contribution, but if we look at IL Generated, we realize that it makes a total difference when we use Lambda Expression in our methods.

### Quik sample:
```csharp
public class Program
{
    static void Main(string[] args)
    {
        FooBar.Multiply(2, 2);

        FooBar.MultiplyExpressionBody(2, 2);
    }
}

public class FooBar
{
    public static int Multiply(int a, int b)
    {
        return a * b;
    }

    public static int MultiplyExpressionBody(int a, int b) => a * b;
}
```

### IL code generated:
**First Sample**
```csharp
.method public hidebysig static int32  Multiply(int32 a,
                                                int32 b) cil managed
{
  // Code size       9 (0x9)
  .maxstack  2
  .locals init ([0] int32 V_0)
  IL_0000:  nop
  IL_0001:  ldarg.0
  IL_0002:  ldarg.1
  IL_0003:  mul
  IL_0004:  stloc.0
  IL_0005:  br.s       IL_0007
  IL_0007:  ldloc.0
  IL_0008:  ret
} // end of method FooBar::Multiply
```

**Second example is much cleaner**
```csharp
.method public hidebysig static int32  MultiplyExpressionBody(int32 a,
                                                              int32 b) cil managed
{
  // Code size       4 (0x4)
  .maxstack  8
  IL_0000:  ldarg.0
  IL_0001:  ldarg.1
  IL_0002:  mul
  IL_0003:  ret
} // end of method FooBar::MultiplyExpressionBody
```

### DIFF
--- a/src/MediatR/Unit.cs
+++ b/src/MediatR/Unit.cs
@@ -1,4 +1,4 @@
-﻿namespace MediatR
+namespace MediatR
 {
     using System;
     using System.Threading.Tasks;
@@ -29,10 +29,7 @@
         ///  - Zero: This object is equal to <paramref name="other" />.
         ///  - Greater than zero: This object is greater than <paramref name="other" />.
         /// </returns>
-        public int CompareTo(Unit other)
-        {
-            return 0;
-        }
+        public int CompareTo(Unit other) => 0;
 
         /// <summary>
         /// Compares the current instance with another object of the same type and returns an integer that indicates whether the current instance precedes, follows, or occurs in the same position in the sort order as the other object.
@@ -45,10 +42,7 @@
         ///  - Zero: This instance occurs in the same position in the sort order as <paramref name="obj" />.
         ///  - Greater than zero: This instance follows <paramref name="obj" /> in the sort order.
         /// </returns>
-        int IComparable.CompareTo(object obj)
-        {
-            return 0;
-        }
+        int IComparable.CompareTo(object obj) => 0;
 
         /// <summary>
         /// Returns a hash code for this instance.
@@ -56,10 +50,7 @@
         /// <returns>
         /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table.
         /// </returns>
-        public override int GetHashCode()
-        {
-            return 0;
-        }
+        public override int GetHashCode() => 0;
 
         /// <summary>
         /// Determines whether the current object is equal to another object of the same type.
@@ -68,10 +59,7 @@
         /// <returns>
         /// <c>true</c> if the current object is equal to the <paramref name="other" /> parameter; otherwise, <c>false</c>.
         /// </returns>
-        public bool Equals(Unit other)
-        {
-            return true;
-        }
+        public bool Equals(Unit other) => true;
 
         /// <summary>
         /// Determines whether the specified <see cref="System.Object" /> is equal to this instance.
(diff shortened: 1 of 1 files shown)


Tornhoof:
I see no difference in https://sharplab.io/#v2:C4LglgNgPgAgDAAhgRgNwFgBQMDMSBMCAwggN5YKW4EIBiA9vQEICGATluZpVXigGwIwAO2AIAsgFcIwMAAcIATwAUIsSwA0Q0QgBGASgo8uPU0gDsCFggBUejNx4BfLEd5JkgtROmyFigFEADzk2AFMAZwiwemEmegATFW9NbTEDBABeAD4rW3ssF0xCoA=

jbogard:
Me neither:

https://gist.github.com/jbogard/9479e77a18b2c6fc23161319aad0220a/revisions#diff-94ec98548a42c8691354944e91e75b1d

pwlsjk:
Difference exists in Debug mode, but not in Release:
https://sharplab.io/#v2:EYLgHgbALANALiAlgGxgExAagD4AEAMABLgIwDcAsAFC4DMxATIQMKEDe1hXdjhAYgHsBAIQCGAJ2ocqXbvVIRCiAHZxCAWQCuyOIgAOyAJ4AKFWtEwlqwsACUnWdNnPiAdkKjCAKhuUZsgF9qBzliEkUzDW1dA0MAUTA9cQBTAGdUxAFlYQE0E0iLKzU7QgBeAD4Pb19qIKpaoA

ralmsdeveloper:
> I see no difference in https://sharplab.io/#v2:C4LglgNgPgAgDAAhgRgNwFgBQMDMSBMCAwggN5YKW4EIBiA9vQEICGATluZpVXigGwIwAO2AIAsgFcIwMAAcIATwAUIsSwA0Q0QgBGASgo8uPU0gDsCFggBUejNx4BfLEd5JkgtROmyFigFEADzk2AFMAZwiwemEmegATFW9NbTEDBABeAD4rW3ssF0xCoA=

@Tornhoof 
IL is very different on the site you used!

image

Use ILDASM and you'll see!

Tornhoof:
@ralmsdeveloper compile your source in Release Mode and try again :)

ralmsdeveloper:
> @ralmsdeveloper compile your source in Release Mode and try again :)

Difference between Debug and Release in this context only applies if I Optimize the Code!

Otherwise, it does not change anything I showed.

image

ralmsdeveloper:
@jbogard  for me besides make the code cleaner, makes it more secure, because if the option of optimization of code is cleared, we guarantee a single IL.

lilasquared:
Cleaner code maybe but unless you’re compiling the source yourself it wouldn’t matter would it? The dlls in nuget I assumed are built in release mode.

ralmsdeveloper:
I gave my little bits here!

If you think this is not important, you can close the issue, I thought very generically, both in Debug/ Release (Optimized) we would have a small gain with it!

ralmsdeveloper:
ping @jbogard.
For me it makes sense to force this optimization, if for you do not, inform me to close the PR.

jbogard:
It’s not an optimization AFAICT, it’s a style change.

I’ve been traveling but can review next week.

Issue #385, opened 2019-03-24
Send common command (query) object and get object response


megafetis:
Pleace add such functionality:

```
object query  = Activator.CreateInstance(requestType);
object response = Mediator.Send(query);
if(response is SomeType resp) 
{
 //...
}
```

jbogard:
Why? What's your use case here?

megafetis:
For sending all requests from frontend and automatic detect Requests and send them.
Then to send a response.

I want to a send all requests from one place (action) without expanding web API controllers, but in busines logic.
In controllers 2 actions: for Queries, and for Commands

megafetis:
For example in Nhibernate ORM (or EF) to get any entity i must know only its id and entityName:
```
object id='1234';
string entityName = 'SomeEntityClass';
object entity = dbSession.get(entityName,id);
```
I want something like this, but not in databases, but in business logic

jbogard:
How would the single controller know how what request type to use, and how would fill it with data? Would the data in the requests come from HTTP request information?

megafetis:
Its simple. From json
The query class will create using Nson.Net lib or manually
```
//js
var httpPayload= {
  __type:'SomeDataQuery',
  id:'1234'
}
axios.get('url',httpPayload).then(resp=>console.log(resp))

```

jbogard:
That's how you instantiate it - but where will the properties on the request come from? JSON.Net? So you'd deserialize from JSON.NET into a request object?

megafetis:
yes using JSON.Net in this case.

jbogard:
I'm not sure about your use case - it sounds a bit like gRPC, and I'm not sure that it's worth all that just to avoid controllers.

But, Publish has an overload that is object-based for a different use case, so it would make sense to create one for `Send`. I'll take a pull request if you want to give this a shot.

megafetis:
Ok Thanks.

lilasquared:
@megafetis we do this today without needing a object based send method.  Change if you change your json object that is deserialized to something like this:

```csharp
public class QueryWrapper
{
    public IRequest Query { get; set; }
}
```

then your json would just look like this: 
```json
{
    "Query": { "__type": "SomeDataQuery", "id": 1234
}
```

since you are using the __type property it will retain the correct type and the object will know it is an IRequest and `.Send(...)` will work just fine.  You may even be able to remove the wrapper layer and simply have your controller accept an IRequest parameter instead of an object.

[...]

PR #415, opened 2019-07-01
Non-generic Send request method

twadrianlis:
This PR tries to resolve issue reported in: https://github.com/jbogard/MediatR/issues/385

by Adding a non generic API for IMediator interface:

```
Task<object> Send(object request, CancellationToken cancellationToken)
```

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -40,6 +40,29 @@ namespace MediatR
             return handler.Handle(request, cancellationToken, _serviceFactory);
         }
 
+        public Task<object> Send(object request, CancellationToken cancellationToken = default)
+        {
+            if(request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+            var requestType = request.GetType();
+            var requestInterfaceType = Array.Find(requestType.GetInterfaces(),
+                i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequest<>) );
+            bool isValidRequest = requestInterfaceType != null;
+            if(isValidRequest)
+            {
+                var responseType = requestInterfaceType.GetGenericArguments()[0];
+                var handler = _requestHandlers.GetOrAdd(requestType,
+                    t => Activator.CreateInstance(typeof(RequestHandlerWrapperImpl<,>).MakeGenericType(requestType, responseType)));
+
+                // call via dynamic dispatch to avoid calling through reflection for performance reasons
+                return (handler as RequestHandlerBase).Handle(request, cancellationToken, _serviceFactory);
+            }
+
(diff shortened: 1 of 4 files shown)


src/MediatR/Mediator.cs
```diff
@@ -40,6 +40,29 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
         }
 
+        public Task<object> Send(object request, CancellationToken cancellationToken = default)
+        {
+            if(request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+            var requestType = request.GetType();
+            var requestInterfaceType = Array.Find(requestType.GetInterfaces(),
```
jbogard: Why not `requestType.GetInterfaces().FirstOrDefault(...)`?

twadrianlis: Sorry for delay, replaced the call with suggestion.

twadrianlis:
@jbogard hello,

Do you have any remarks or feedback I could implement here?

remcoros:
I have done this in the past with this snippet:

```csharp
using System.Threading.Tasks;

namespace MediatR
{
    public static class MediatorExtensions
    {
        public static async Task<object> Send(this IMediator mediator, object request)
        {
            return await mediator.Send((dynamic)request);
        }
    }
}
```

requires you to add Microsoft.CSharp reference (or nuget package in case of netstandard) to your project.

I have not tested anything performance wise, but found it cleaner than reflection.

Just tried it, and still works (used your unit test).

src/MediatR/Mediator.cs
```diff
@@ -40,6 +41,29 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
             return handler.Handle(request, cancellationToken, _serviceFactory);
         }
 
+        public Task<object> Send(object request, CancellationToken cancellationToken = default)
+        {
+            if(request == null)
```
crnd: Write the ifs with an empty space so the code style stays coherent. So
`if(request == null)` to `if (request == null)` and the same thing for the if below.

src/MediatR/Mediator.cs
```diff
@@ -40,6 +41,29 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
+        public Task<object> Send(object request, CancellationToken cancellationToken = default)
+        {
+            if(request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+            var requestType = request.GetType();
+            var requestInterfaceType = requestType.GetInterfaces().FirstOrDefault(
+                i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequest<>) );
+            bool isValidRequest = requestInterfaceType != null;
```
crnd: Is there a reason for using `bool` here instead of just going with `var`?

Issue #429, opened 2019-08-17
Provide the ability to get the notification object in PublishCore


ahocquet:
I've developed a `Polly` based custom Mediator implementation, in order to add exponential retries and fallback. Therefore, it never throws an exception to the `Publish` caller.

The overriden PublishCore is clean and simple : 
```
protected override async Task PublishCore(IEnumerable<Func<Task>> allHandlers)
        {
            var tasks = allHandlers.Select(t => _policy.ExecuteAsync(t));
            await Task.WhenAll(tasks).ConfigureAwait(false);
        }
```

The Polly policy is defined in the constructor, but it could also be injected. 

When my `Fallback` policy triggers, I would like to send the failed notification to an Azure Service Bus Queue, so I can monitor it and replay it later through an Azure Function for example.

In order to achieve that, I would need to get access to the underlying `INotification` object in the `PublishCore` method. I see two possibilities:
- Inject a `IEnumerable<Expression<Func<Task>>>` instead of a `IEnumerable<Func<Task>>`. Thanks to reflection, I would get access to the `INotification` object.
- Introduce a breaking change by updating the `PublishCore` signature in order to provide an `INotification notification` parameter, even though it wouldn't be used in the default implementation.

If you're ok with the idea, I can send a PR based on the preferred solution.

ahocquet:
I could get it done by implementing this logic in a handler decorator instead of a custom Mediator implementation. It works great, so I don't know if my request is still relevant. 

Maybe it's better to keep this logic apart, and maintaining the single responsibility principle. But on the other side, it comes with some plumbing that not all containers support (Asp.net core container, for example, doesn't have decorators support out of the box).

bartosz6:
#353 linked issue

I think a decorator is clean a good solution. No need to modify the mediatr.

ahocquet:
Very well then, let's keep it that way.

jbogard:
FYI this next release will have a breaking change anyway, so if we want to introduce a breaking change, now would be the time.

@bartosz6 I can see decorators being a possibility - but one of the reasons I added the whole behavior pipeline was because of sparse support in DI containers for decorators.

ahocquet:
@jbogard If you think this feature makes sense, I'll be pleased to send you a PR. :smile:

jbogard:
I think it's reasonable - people have asked for behaviors for notifications but to me that never made much sense, because some folks want behaviors around individual handlers vs. ALL the handlers.

[...]

PR #431, opened 2019-08-19
Add INotification and CancellationToken to PublishCore

ahocquet:
Related to #429 

This PR adds `INotification` and `CancellationToken` parameters to the `PublishCore` method. It opens  new possibilities for custom `Mediator` implementations.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <Authors>Jimmy Bogard</Authors>
     <LangVersion>latest</LangVersion>
-    <VersionPrefix>7.0.0</VersionPrefix>
+    <VersionPrefix>8.0.0</VersionPrefix>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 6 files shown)


ahocquet:
I haven't included any `Polly` example, but if you think it makes sense then I can add some. Even though a real `INotification` pipe implementation would better a fit for this kind of stuff.

valin07:
Hi! Any plans on releasing these changes to nuget soon? Still seeing v7.0.0.

casually-creative:
How does one use this change to implement behaviors for INotifications? I'm implementing domain events with the help of notifications, and I want to persist every domain event to a database through a middleware/pipeline. I read this was initially not supported but this change would be a first step, but I'm struggling with implementing this, I can't find any other documentation about this. Any help or code samples would be appreciated, thanks!

CasperAlant:
@casually-creative Hello! I have the same use case. Were you able to make any progress?

casually-creative:
> @casually-creative Hello! I have the same use case. Were you able to make any progress?

Heya! No, unfortunately not. I found it too hard to understand how to create a custom MediatR implementation, and thought it wasn't worth it to achieve what I wanted. I ended up dropping the idea of persisting domain events with a notification behavior pipeline. I implemented a custom service that I call after doing my .SaveChangesAsync() call.

PR #432, opened 2019-08-20
Support for sending a request to many handlers

vmikhailov:
Hi. It would be great to have ability to send a request to multiple handlers and receive result back. Please consider potential implementation of the feature.

### DIFF
--- a/samples/MediatR.Examples.AspNetCore/Program.cs
+++ b/samples/MediatR.Examples.AspNetCore/Program.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using MediatR.Pipeline;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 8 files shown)


jbogard:
What's the use case here? A handler could just as easily do this behind the scenes. What about the pipeline behavior? This assumes a single pipeline across *all* handlers, that it can't execute per handler. What about `IAsyncEnumerable`?

vmikhailov:
We have sort of in-proc message processing and we need to process some of messages transactionally and some of them are just notifications. Handlers for transactional processing does not know about each other and i potentially could have a handler for collection but it's essentially the same abstraction what MediatR does.
I've intentionally made single pipeline because i think there should be single pre/post handler for a group. 
Regarding IAsyncEnumarable - I don't really use the concept so far. If you think it is usable here then i can do. For Task<IEnumerable<T>> does pretty much the same job because i don't need syntax sugar around it. (or i misunderstanding the concept)

vmikhailov:
Hey, @jbogard, do you still have concerns about the feature? it is holding us a bit.

jbogard:
This isn't something I'm planning on merging soon. It's a significant departure from existing behavior, and not something our teams would use at the moment. It's going to take a bit more discussion to describe and capture the appropriate behavior, that would likely result in a different request/handler/pipeline types.

jbogard:
Closing for now. We can discuss in an issue instead

PR #468, opened 2019-12-07
The exception handler improvements

Shumigaj:
The exception handler has been modified to work like try/catch with the ability to override inherited exception handlers


Shumigaj:
I decided that using inheritance is better/clear to override the handler than overriding by the name of the handler (3e5d4c5 vs 49b32a1)

Shumigaj:
Added additional behavior to cover situations where we can't handle the exception and just want need to perform some actions (e.g. logging). It will also be useful to log all unhandled exceptions.

The main difference from the exception handler is that all actions for the exception will be called (exception handlers will be executed while the exception remains unhandled).

jbogard:
Looks good. After this, can you update the wiki?

Shumigaj:
Yes, I'll update the wiki

PR #487, opened 2020-01-23
composite requests

jasonHooten:
**TL:DR**
A way to get many responses in a single 'request' without duplicating code.

**Use Case**
Admittedly this is a edge case but I was working on a GraphQL server and needed to cache (locally via Apollo) lots of data from multiple endpoints on initial page load. I didn't want to make a new IRequest that duplicated other end points so I consolidated them together under one request/handler and returned aggregate as a dynamic object (ExpandoObject) to easily be serialized in to JSON. The basic use case looks something this...

```c#
var mediator = container.GetInstance<IMediator>();

var pongResponse = await mediator.Send(new Ping());
var fooResponse = await mediator.Send(new Foo());

dynamic response = await mediator.Send(new AsyncCompositeRequest(
      new List<IBaseRequest> { new Ping(), new Foo() } 
));
Assert.Equal(response.Ping.Message, "Ping Pong");
Assert.Equal(response.Foo.Total, 9001);
```
where it uses the request names to group responses together. The other options being to provide a a dictionary instead...

```c#
var mediator = container.GetInstance<IMediator>();

var pongResponse = await mediator.Send(new Ping());
var fooResponse = await mediator.Send(new Foo());

dynamic response = await mediator.Send(new AsyncCompositeRequest(
      new Dictionary<string, IBaseRequest>()
      {
            { "AAA", new Ping() },
            { "BBB",  new Foo()  }
       }
));
Assert.Equal(response.AAA.Message, "Ping Pong");
Assert.Equal(response.BBB.Total, 9001);
```

The only real issue I have is the need to add AsyncCompositeRequestHandler to the DI container since it doesn't already look in the MediatR library.

```c#
scanner.AssemblyContainingType(typeof(AsyncCompositeRequestHandler));
scanner.IncludeNamespaceContainingType<AsyncCompositeRequestHandler>();
```

This very well may be outside the scope of of what you want MediatR to be but I found it useful so I figured I share. 

Thanks!


lilasquared:
@jasonHooten this is pretty cool, but what prevents you from doing this on the client side by hitting the individual endpoints and caching the response?  What is the value of moving this to the server and aggregating into one endpoint?

joaomatossilva:
For mobile applications, it's different to make one single call rather than multiple ones. Less DNS and connections handshakes. (not saying this is the case, just giving out a possible use case for this)

lilasquared:
Oh yeah didn't think of mobile.  I think it would be relatively simple to make a helper to do this rather than it be its own new handler.  I'm wondering what the expectation is for failures as well as pipeline behaviors.  With nested handlers you end up hitting pipelines twice and such.

jasonHooten:
@lilasquared fair point on exception pipelines. My uses cases/tests have been fairly benign so I'd have to play around with some of the pipeline behaviors before coming up with a solution. I kept it as a handler since I'm rolling my own simple GraphQL router and wanted to keep the design paradigm consistent between single requests and batched requests. 

That said I could be swayed either way since this is an uncommon operation and may have difficulties, as you pointed out, getting the error back to the user.

@JosephWoodward you're correct, I have about ~7 endpoints worth of data I want to preload and I'd like to have as little chatter as possible.

src/MediatR/Composite/CompositeRequestHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            _mediator = mediator;
+        }
+
+        public Task<ExpandoObject> Handle(CompositeRequest compositeRequest, CancellationToken cancellationToken)
+        {
+            IDictionary<string, object> response = new ExpandoObject();
+            foreach (var request in compositeRequest.Requests)
+            {
+                response[request.Key] = _mediator.Send(request.Value);
```
lilasquared: I'm not sure I understand the point of the `CompositeRequestHandler`.  All mediator handlers are async by default, so in this case aren't you just storing the tasks in the `response[request.Key]`, which are going to continue running as your loop through the requests?  In your tests you are only using `AsyncCompositeRequest`.

jasonHooten: No you're 100% correct. That was a hold over from an attempting synchronous method that never worked like I wanted it to (was to be inline the synchronous RequestHandler). I'll remove it.

jasonHooten:
Played around with the pipeline exception handlers, things seem to be working as expected. 

```c#
             dynamic response = await mediator.Send(new AsyncCompositeRequest(new List<IBaseRequest>
                {
                    new Ping { Message = "Ping" },
                    new Foo()
                }
            ));

            Assert.Equal("Ping Thrown Handled by Type", response.Ping.Message);
            Assert.Equal(9001, response.Foo.Total);
```

and also seems to work with the unhandled exception.

```c#
            var compositeRequest = new AsyncCompositeRequest(new List<IBaseRequest>
                {
                    new Ping { Message = "Ping" },
                    new Foo()
                }
            );
            await Should.ThrowAsync<PingException>(async () =>
            {
                await mediator.Send(compositeRequest);
            });

            Assert.Equal("Ping Thrown Not Handled", (compositeRequest.Requests["Ping"] as Ping).Message);
```
Is there anything I'm missing or should test differently?

jbogard:
I think when I've done this in the past, I made up my own "internal" request types as like a chain of responsibility. Like a search screen that has different strategies for each set of fields

jasonHooten:
I'll try and put together a demo of where this is useful or some interesting applications.

jasonHooten:
I'm actually going to shut this pull request down, maybe make it an extension eventually. I'm pretty sure this is outside of scope of MediatR and would need more tooling to be widely useful. 

Thanks

vitalybibikov:
I have implemented this thing some time ago as a set of extensions.
As I can see it won't be included in the repo, at least, for now.

@jasonHooten  You are free to comment in the repo, I guess.

https://github.com/vitalybibikov/MediatR.Extensions.UnitOfWork

PR #495, opened 2020-02-17
Improvement of notification handlers

VitalyKorolyov:
The pull request contains improvement of notification handlers behavior. Now you can apply notification handler's order value the way you want.

### DIFF
--- a/src/MediatR/Internal/NotificationHandlerWrapper.cs
+++ b/src/MediatR/Internal/NotificationHandlerWrapper.cs
@@ -26,10 +26,10 @@ namespace MediatR.Internal
 
             IEnumerable<Func<INotification, CancellationToken, Task>> handlersToPublish;
 
-            if (allHandlers.All(x => x.GetType().GetCustomAttribute<NotificationOrderAttribute>() != null))
+            if (allHandlers.All(x => x.GetType().GetCustomAttribute<NotificationHandlerOrderAttribute>() != null))
             {
                 var groupedHandlers = allHandlers
-                    .GroupBy(x => x.GetType().GetCustomAttribute<NotificationOrderAttribute>().Value)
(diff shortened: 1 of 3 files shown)


jbogard:
A couple of things:

 - In general, notification handlers shouldn't care about order of each other. They should be independent
 - Containers can support ordered registration - why not lean on those?

VitalyKorolyov:
I agree, but containers don't allow you to manage processing of notification handlers. For example, we have 3 notification handlers and first one should be processed at the beginning then others should be processed in parallel. I found (https://github.com/jbogard/MediatR/tree/master/samples/MediatR.Examples.PublishStrategies) how to process **all** notification handlers in parallel. My pull request allows to customize process of notification handlers.

src/MediatR/Internal/NotificationHandlerWrapper.cs
```diff
@@ -18,11 +20,35 @@ internal class NotificationHandlerWrapperImpl<TNotification> : NotificationHandl
...
+                    .GroupBy(x => x.GetType().GetCustomAttribute<NotificationHandlerOrderAttribute>().Value)
+                    .OrderBy(x => x.Key);
+
+                foreach (var handlersGroup in groupedHandlers)
+                {
+                    handlersToPublish = handlersGroup.Select(x =>
+                        new Func<INotification, CancellationToken, Task>((theNotification, theToken) =>
+                            x.Handle((TNotification) theNotification, theToken)));
+
+                    publish(handlersToPublish, notification, cancellationToken);
```
VitalyKorolyov: I suppose I missed await.

jbogard:
Why do you need a specific order? The point of notification handlers is that they're isolated from one another.

You can also add this example to the wiki in publish strategies.

jbogard:
I don't think this fits in the spirit of the core of MediatR. You can add this example to the wiki, however.

VitalyKorolyov:
ok, thank you.

PR #509, opened 2020-04-12
Pipeline Priority Attribute to control the order of middleware execution

dasiths:
The current MediatR pipeline relies on the order which individual `IPipelineBehavior` registrations that are registered in the DI container.

The new `PipelinePriorityAttribute` allows for this behaviour to extended. Any classes marked with this attribute will get ordered with their specified value while still honoring the original order of their registration for pipeline behaviours of the same priority order.

We can now ensure the following order regardless of how things are registered as an example of using this pattern.

```csharp
            services.AddTransient(typeof(IPipelineBehavior<,>), typeof(RequestPreProcessorBehavior<,>));
            services.AddTransient(typeof(IPipelineBehavior<,>), typeof(RequestPostProcessorBehavior<,>));
            services.AddTransient(typeof(IPipelineBehavior<,>), typeof(RequestExceptionActionProcessorBehavior<,>));
            services.AddTransient(typeof(IPipelineBehavior<,>), typeof(RequestExceptionProcessorBehavior<,>));
```

Furthermore, this order is now not reliant of the DI container implementation of `IEnumerable<IPipelineBehavior<,>>` to return in the order they were registered. The pipeline implementations can be in control of their own order/priority as required.

Couple of small use cases I could think of:

- This makes it easier to scan for open generic types that inherit from `IPipelineBehavior` and just register them.
- Ability to register multiple high priority middleware like Logging with insurance that everything is captured inside its Handle method regardless of how it is registered.

It doesn't break any current behaviour and only has a very small performance hit due to the extra check of the order. I was inspired by the `Order` property of https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.apiexplorer.iapidescriptionprovider?view=aspnetcore-3.1 but the Ascending/Descending behvaiour might be a bit diffrent here. I'm happy to change it if required.

This is my first PR to a major community project and I'm not sure I'm following the guidelines here. Sorry in advance. What do you think? Would love some feedback.


dasiths:
Not exactly sure why the build failed. I can see this "C:\Users\appveyor\.nuget\packages\microsoft.build.tasks.git\1.0.0-beta-62925-02\build\Microsoft.Build.Tasks.Git.targets(20,5): error MSB4018: The "Microsoft.Build.Tasks.Git.LocateRepository" task failed unexpectedly. ". Could it be because I created the PR from a fork?

rus-art:
The same build fail in my PR #518

src/MediatR/Internal/RequestHandlerWrapper.cs
```diff
@@ -1,3 +1,5 @@
+using MediatR.Pipeline;
```
no1melman, line 1: Using in a different place?

src/MediatR/Internal/RequestHandlerWrapper.cs
```diff
@@ -64,7 +66,11 @@ internal class RequestHandlerWrapperImpl<TRequest, TResponse> : RequestHandlerWr
             return serviceFactory
                 .GetInstances<IPipelineBehavior<TRequest, TResponse>>()
                 .Reverse()
-                .Aggregate((RequestHandlerDelegate<TResponse>) Handler, (next, pipeline) => () => pipeline.Handle((TRequest)request, cancellationToken, next))();
+                .OrderBy(i => i.GetType()
+                                  .GetCustomAttributes(typeof(PipelinePriorityAttribute), true)
```
no1melman, line 70: So are you suggesting the only way to get priority is through an Attribute? not adding it to an interface?

src/MediatR/Pipeline/PipelinePriorityAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+            Priority = priority;
+        }
+
+        /// <summary>
+        /// Priority of execution. Higher the the number the higher the priority. See <see cref="PipelinePriorityOrder"/> for predefined priorities.
+        /// </summary>
+        public int Priority { get; set; }
+    }
+
+    public static class PipelinePriorityOrder
```
no1melman, line 22: Should we move this out into its own file?

src/MediatR/Internal/RequestHandlerWrapper.cs
```diff
@@ -64,7 +66,11 @@ internal class RequestHandlerWrapperImpl<TRequest, TResponse> : RequestHandlerWr
             return serviceFactory
                 .GetInstances<IPipelineBehavior<TRequest, TResponse>>()
                 .Reverse()
-                .Aggregate((RequestHandlerDelegate<TResponse>) Handler, (next, pipeline) => () => pipeline.Handle((TRequest)request, cancellationToken, next))();
+                .OrderBy(i => i.GetType()
```
no1melman, line 69: How do we keep the order consistent, do we care so much? Just wondering if something has the same priority as something else, it could keep jumping around in execution order - this may not be a problem because the priorities are the same - but just thinking about it being a possible nuance that people rely on too heavily thinking they get a guaranteed order of execution.

src/MediatR/Internal/RequestHandlerWrapper.cs
dasiths: > So are you suggesting the only way to get priority is through an Attribute? got any thoughts on adding it to an interface?

I thought about doing it via an interface but then there is no way to override. (i.e. In an inheritance scenario if the child class sets a different priority to the parent. With an attribute the parent's one get overridden.)

The priority attribute is optional so it doesn't break any existing code and honors the order of registration in the DI container.

> I wouldn't mind seeing some debug function we could call to print out the priority along with type name and full name, like you could do with structuremaps WhatDoIHave, WhatDidIScan - this just may help with debugging stuff...

I can certainly do that but we would need to do that per `request type` because that's what's how MediatR resolves `IRequestHandler<TRequest, TResult>` and middleware.

src/MediatR/Internal/RequestHandlerWrapper.cs
dasiths: If we have middleware the order of execution does matter a lot. They can enrich the request or context with information the lower order middleware depend on.

The `OrderBy` function is an implementation of [stable sort](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderby?redirectedfrom=MSDN&view=netcore-3.1#System_Linq_Enumerable_OrderBy__2_System_Collections_Generic_IEnumerable___0__System_Func___0___1__). So it won't jump around for the things that are of the same order. 

This implementation doesn't get in the way of anyone who doesn't want to use it but helps a developer guarantee the high/low priority order of something like a logging or exception handling middleware.

src/MediatR/Pipeline/PipelinePriorityAttribute.cs
dasiths: We certainly can. Do you want to keep it in the same namespace?

src/MediatR/Internal/RequestHandlerWrapper.cs
dasiths: This is due to us referencing `PipelinePriorityAttribute`. Do you prefer it be in a different namespace? Remember we can't put it in the  `Internal` namespace because it needs to be used by others as well.

src/MediatR/Pipeline/PipelinePriorityAttribute.cs
no1melman: I think it is sensible to keep it in the same namespace, otherwise I would move all priority stuff into a `MediatR.Priority` namespace for neatness... but I think that is a community call

dasiths:
@jbogard thoughts?

BalassaMarton:
IMHO this just adds complexity and actually reduces predictability. Behaviors are usually registered manually, in a specific order, so you only have to look at one place in your code to see how your pipeline is set up. If you still want to control this by an attribute, I suggest you write an extension that registers the behaviors in the order specified by the attributes.

dasiths:
> IMHO this just adds complexity and actually reduces predictability. Behaviors are usually registered manually, in a specific order, so you only have to look at one place in your code to see how your pipeline is set up. If you still want to control this by an attribute, I suggest you write an extension that registers the behaviors in the order specified by the attributes.

First of all thank you for your input. I do agree that it adds a bit of complexity but it is my opinion that this feature gives MediatR the ability to be more DI provider agnostic. I think the PR strikes a good middle ground.

- It doesn't affect any existing order of registration or for people who don't want to use the attribute. Business as usual. In fact the decision to use an attribute for this was driven from not wanting to interfere with existing use cases. It doesn't change the predictability for anyone who doesn't want to use the attribute.
- You assume that all DI implementations give back an `IEnumerable<IPipelineBehaviour<T>>` in order of the registration. If we are being technical `IEnumerable` does not have to guarantee any order (https://stackoverflow.com/a/10409943/422427) hence it's mostly a nice to have even in the DI containers that do support it. By using the attribute introduced here a developer can take control of the order of exactly how they will be executed and gives MediatR more power to work with any DI provider.
- Therefore it doesn't make a lot of sense to have another extension method to register these as the order is enforced during the runtime and not register time for *core parts of MediatR* with this change. This PR would would only affect a developer if they were registering the core pieces of Mediator in a different order to how the default extension method does (highly unlikely as it would break the functionality of MediatR anyway).

Thank again for taking the time to review the PR :)

shashanktomar:
What happened to this, it is a very handy feature to have.

jbogard:
I'm open to the idea, but I'd rather not couple to attributes.

I'd rather take the approach of ASP.NET Core middleware registration, where you EXPLICITLY register middleware in the EXACT order you want it to execute. I don't want to support scanning of middleware, that's too confusing, and having the order specified on the type itself rather than a central registration point means you can't see in one place the exact order in which your behaviors will run.

SilentBlueD666:
I agree, it much easier to control and manage the order via DI registration order in one place.

dasiths:
> I'm open to the idea, but I'd rather not couple to attributes.
> 
> I'd rather take the approach of ASP.NET Core middleware registration, where you EXPLICITLY register middleware in the EXACT order you want it to execute. I don't want to support scanning of middleware, that's too confusing, and having the order specified on the type itself rather than a central registration point means you can't see in one place the exact order in which your behaviors will run.

@jbogard if we want to go with that approach I'm thinking of something like this...

```csharp
public interface IPipelineBehaviorBuilder {
 public void Register (Type T);
 public void IReadOnlyList<IPipelineBehavior<TRequest, TResponse>> GetPipelineBehaviors<TRequest, TRespons>();
}
```
- A type for building the middleware pipeline `PipelineBehaviorBuilder` which you call to register your Behaviour like `builder.Register(typeof(RequestPreProcessorBehavior<,>))`.
- We use the serviceFactory inside the `PipelineBehaviorBuilder` to resolve the instances via DI but we control the order they are returned. We don't use the serviceFactory in the `RequestHandlerWrapper`. Instead we call `PipelineBahaviourBuilder.GetPipelineBehaviors<TRequest, TResponse>()`

How does that sound?

dasiths:
> I agree, it much easier to control and manage the order via DI registration order in one place.

See my reply above about why we can't rely on DI alone.

> You assume that all DI implementations give back an IEnumerable<IPipelineBehaviour<T>> in order of the registration. If we are being technical IEnumerable does not have to guarantee any order (https://stackoverflow.com/a/10409943/422427) hence it's mostly a nice to have even in the DI containers that do support it. By using the attribute introduced here a developer can take control of the order of exactly how they will be executed and gives MediatR more power to work with any DI provider.

jbogard:
Right, you basically would need to sidestep the container's behavior for `IEnumerable<>` completely. I was **trying** to avoid this because this puts more responsibility on MediatR's shoulders. The container will also be able to handle generic constraints for us.

dasiths:
> Right, you basically would need to sidestep the container's behavior for `IEnumerable<>` completely. I was **trying** to avoid this because this puts more responsibility on MediatR's shoulders. The container will also be able to handle generic constraints for us.

So I'm thinking of still leveraging the DI container for it. But then reordering the `IEnumerable<>` to enforce the registered behaviour. This only puts the ordering logic on the shoulders of MediatR. 

We would still require some changes to the container specific Mediator extensions to support the new `IPipelineBehaviorBuilder` type and registering behaviours on it but I'm assuming you have done that type of thing before and can guide me with best approach.

dasiths:
@jbogard rather than push the change here I created a draft of how it would potentially look here https://github.com/dasiths/MediatR/pull/2 (It's a PR targeting my fork)

It satisfies your requirement of explicitly registering the order and also doesn't break any existing behaviour. If no instance of `IBehaviourOrder` is registered it falls back to using the `IEnumerable<,>` from ServiceFactory just as before.

```csharp
// registering example
                var x = BehaviorOrderBuilder
                    .Create(t => services.AddTransient(typeof(IPipelineBehavior<,>), t))
                    .RegisterBehaviour(typeof(OuterBehavior<,>))
                    .RegisterBehaviour(typeof(InnerBehavior<,>))
                    .RegisterBehaviour(typeof(ConstrainedBehavior<,>));
                
                services.AddSingleton<IBehaviorOrder>(x);
```

Please let me know what you think? If you're happy with this approach I'll push the changes here and update this PR.

remcoros:
This adds a lot of (reflection, sorting) logic for every request. You need to consider that all the 'Handle' logic is called for every http request (in web workloads) and will be called A LOT.

I'd prefer to keep the direct usage of the container.

Also, unless you provide a custom IBehaviorOrder, there is no difference between using services.Add(..) calls in order vs BehaviorOrderBuilder.Register, just a different way of writing it.

If adding behavior order based on attributes is what you are after (I presume by implementing an AttributeBasedBehaviorOrder), I think this is best done by an extension method on top of IServiceCollection / the container.

dasiths:
> This adds a lot of (reflection, sorting) logic for every request. You need to consider that all the 'Handle' logic is called for every http request (in web workloads) and will be called A LOT.
> 
> I'd prefer to keep the direct usage of the container.
> 
> Also, unless you provide a custom IBehaviorOrder, there is no difference between using services.Add(..) calls in order vs BehaviorOrderBuilder.Register, just a different way of writing it.
> 
> If adding behavior order based on attributes is what you are after (I presume by implementing an AttributeBasedBehaviorOrder), I think this is best done by an extension method on top of IServiceCollection / the container.

To be clear, the approach of using attributes and the approach of using `IBehaviorOrder` are mutually exclusive. I did a POC for using `IBehaviorOrder` because @jbogard suggested he preferred a way to directly register them in order required.

So if you want to **preserve the order** in a container agnostic way, then you can't use the container implementation alone for it. See my comments above. Basically you can't **just** register `IPipelineBhaviour<,>` in the container, the order need to be captured externally.

If you want to use Attributes then you don't need to side step the `IEnumerable<>` implementation of the container. (i.e. No changes to registration in container or resolving `IPipelineBhaviour<,>`. Attributes can be inspected and ordered inside the `RequestHandlerWrapper`.

Again, the two approaches are mutually exclusive.

To address the reflection performance issues, I would consider a feature switch which doesn't try to resolve a `IBehaviorOrder` from the container which would then simply use the native container implementation of `IEnuimerable<>`. But if you want to control explicit ordering (i.e. use `IBehaviorOrder`) then you can't side step having to use some sort of reflection IMO. We can certainly optimize the code I've currently got. But the aim here is to understand if it's a viable option.

I agree there are no elegant solution here to all 
A. Let container handler everything.
B. Enforce order.
C. Register in one place.

It will be a matter of compromise.

dadhi:
Sorry for bursting into the discussion.
Just adding to the info, that some DI containers support [collection order customization](https://github.com/dadhi/DryIoc/blob/master/docs/DryIoc.Docs/Wrappers.md#custom-order-of-items-in-the-collection-wrapper), so you  may utilize the DI tools to the full extent.

dasiths:
> Sorry for bursting into the discussion.
> Just adding to the info, that some DI containers support [collection order customization](https://github.com/dadhi/DryIoc/blob/master/docs/DryIoc.Docs/Wrappers.md#custom-order-of-items-in-the-collection-wrapper), so you may utilize the DI tools to the full extent.

Yes conceptually the original attribute based ordering is a similar mechanism for metadata based ordering like in DryLoc. 

The aim is to not depend on DI Implementation because not every DI supports it. Like @jbogard said we need to sidestep the DI implementation to do it (if you want to do without attributes). Just waiting to see what his preferred way is (or if this feature is worthwhile at all).

We still need to pick 2 out of the 3 here IMO
> A. Let container handler everything.
B. Enforce order in an DI agnostic way.
C. Register in one place.

remcoros:
Which DI container(s) do not guarantee the order of registration?

dasiths:
> Which DI container(s) do not guarantee the order of registration?

Thanks for responding. Anything that resolves `IEnumerable<T>` by definition does not need to support ordering. Autofac didn't support it a while back (https://stackoverflow.com/questions/5218803/is-order-of-dependencies-guaranteed-when-injecting-ienumerablet). Even `Microsoft.Extensions.DependencyInjection` didn't for a while if I remember correctly (I remember MediatR MS DI extension had a try catch logic to support `IEnumarable<T>`).

**But** most DI frameworks currently do support it. This is where it gets muddy...

The point here is that we are building on the promise that a MediatR request goes through the pipeline in the order we register the pipeline behaviours in. But that's **brittle** because we are asking for a  `IEnumerable<T>`. It's like executing a SQL query on a database (without an order by clause) and expecting the result to be in order each time because the driver/library orders it implicitly. But I totally get the fact that this might be over thinking on my part and expecting a level of preciseness not warranted.

On that note, it would be precise if we relied on a `IOrderedEnumerable<TElement>` and then left everything up to the DI. Containers that implement that could guarantee order each time it's resolved.

jbogard:
For this, I'd rather NOT rely on the container for ALL cases versus DO rely on the container for SOME cases. The better way to approach this is to move registration for pipeline behaviors into the DI package where it will preserve the order for you (or provide some way of specifying order).

PR #518, opened 2020-05-17
Enable Nullable Reference Types

### LINKED ISSUE #503: Enable Nullable Reference Types
Nullable Reference Types were introduced in C# 8.

In projects and libraries already using Nullable Reference Types depending on MediatR it would be very useful to have it. It provides perfect integration and avoids a lot of problems.

rus-art:
Hi! This PR resolves #503

Added target framework netstandard2.1

I've assumed that:
- request cannot be null in any way
- response can be null
- RequestExceptionHandlerState cannot be handled with response of null if generic parameter TResponse is declared as non-nullable

The source of file NullableAttributes.cs is this [article](https://www.meziantou.net/how-to-use-nullable-reference-types-in-dotnet-standard-2-0-and-dotnet-.htm#add-nullable-attribu)

### DIFF
--- a/src/MediatR/IMediator.cs
+++ b/src/MediatR/IMediator.cs
@@ -24,7 +24,7 @@ namespace MediatR
         /// <param name="request">Request object</param>
         /// <param name="cancellationToken">Optional cancellation token</param>
         /// <returns>A task that represents the send operation. The task result contains the type erased handler response</returns>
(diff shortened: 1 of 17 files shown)


lilasquared:
Can you checkout `Mediator.cs`? Something might be wrong with your line endings, it is showing a full file diff.

rus-art:
@lilasquared Thanks, it's ok now

PR #529, opened 2020-06-28
simplify MinVer reference

adamralph:
Unfortunately the tooling prefers the verbose variant, but this terse one is enough.

### DIFF
--- a/src/MediatR/MediatR.csproj
+++ b/src/MediatR/MediatR.csproj
@@ -28,9 +28,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="MinVer" Version="2.3.0">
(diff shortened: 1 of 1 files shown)


adamralph:
Regarding the build failure, it seems https://github.com/jbogard/MediatR/blob/22014728fcd0122c5bd2f314831d65687eb859d4/Push.ps1#L4 doesn't work as intended. I suspect `$Env:MYGET_MEDIATR_CI_API_KEY` is empty rather than `null`.

adamralph:
The build problem is fixed in https://github.com/jbogard/MediatR/pull/530.

adamralph:
Looks like `master` is broken now.

jbogard:
Thanks! I broke something else along the way...

PR #552, opened 2020-09-17
Defer handler reflection

SimonCropp:
no point doing the reflection on every send. only do it when _requestHandlers does not already contain a cached instance

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -58,7 +58,7 @@ namespace MediatR
 
                     if (!isValidRequest)
                     {
-                        throw new ArgumentException($"{nameof(request)} does not implement ${nameof(IRequest)}");
+                        throw new ArgumentException($"{requestType.Name} does not implement {nameof(IRequest)}");
                     }
 
                     var responseType = requestInterfaceType!.GetGenericArguments()[0];
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -49,9 +49,9 @@ namespace MediatR
             }
             var requestType = request.GetType();
             var handler = _requestHandlers.GetOrAdd(requestType,
-                t =>
+                requestTypeKey =>
                 {
-                    var requestInterfaceType = requestType
+                    var requestInterfaceType = requestTypeKey
                         .GetInterfaces()
                         .FirstOrDefault(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequest<>));
                     var isValidRequest = requestInterfaceType != null;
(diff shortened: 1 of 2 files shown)


src/MediatR/Mediator.cs
```diff
@@ -48,19 +48,22 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
-            {
-                throw new ArgumentException($"{nameof(request)} does not implement ${nameof(IRequest)}");
-            }
-
-            var responseType = requestInterfaceType!.GetGenericArguments()[0];
             var handler = _requestHandlers.GetOrAdd(requestType,
-                t => Activator.CreateInstance(typeof(RequestHandlerWrapperImpl<,>).MakeGenericType(requestType, responseType)));
+                t =>
+                {
+                    var requestInterfaceType = requestType
```
jbogard, line 54: Should this use `t` to avoid allocations?

SimonCropp: good catch. fixed

src/MediatR/Mediator.cs
```diff
@@ -48,19 +48,22 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
...
+                requestTypeKey =>
+                {
+                    var requestInterfaceType = requestTypeKey
+                        .GetInterfaces()
+                        .FirstOrDefault(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequest<>));
+                    var isValidRequest = requestInterfaceType != null;
+
+                    if (!isValidRequest)
+                    {
+                        throw new ArgumentException($"{nameof(request)} does not implement ${nameof(IRequest)}");
```
no1melman, line 61: Moving this here - will it cause any issues?

jbogard: Stack trace would be different?

no1melman: Yeah that would be part of it. Which may lead it to be hidden?

jbogard: Hidden by what? Are you suggesting the exception will get swallowed somehow?

SimonCropp: i added a test for that scenario

src/MediatR/Mediator.cs
```diff
@@ -58,7 +58,7 @@ public Task<TResponse> Send<TResponse>(IRequest<TResponse> request, Cancellation
 
                     if (!isValidRequest)
                     {
-                        throw new ArgumentException($"{nameof(request)} does not implement ${nameof(IRequest)}");
+                        throw new ArgumentException($"{requestType.Name} does not implement {nameof(IRequest)}");
```
SimonCropp, line 61: i also found a bug with the exception message

jbogard: Should we add the argument name here? `nameof(requestType)`?

SimonCropp: done

Issue #371, opened 2019-02-22
Async method naming conventions


jamesharling:
Apologies if this has been asked before, I couldn't find an issue about it.

C#/.NET design docs recommend the use of the `Async` suffix for async methods, and I think this is a well-established convention. The MediatR interfaces however don't use this suffix - e.g. `Task<TResponse> Handle(TRequest request, CancellationToken cancellationToken)`.

Is there any plan to align with this convention?

jbogard:
No. Otherwise every single method would have Async in its name. So I opted for what some other, pure-async libraries do:

https://docs.particular.net/nservicebus/upgrades/5to6/async-suffix

mwarrier:
Disagree with this, but accept the answer. For the .NET library I think having Async makes perfect sense.

PR #560, opened 2020-09-22
Added async suffix in asynchronous methods  #371

souzara:
Hi, 

This pull request, is adding suffix async in asynchronous methods, and it's ref.  #371 Issue.

I read nservicebus case(about async suffix), and I understand about It does not use async suffix, but, in the real world, the async suffix help a lot all developers to avoid some simple bugs, otherwise, developers should back to code and fix it, and if we add this suffix, we'll help a lot of people.

In this pull request I renamed all asynchronous methods.

Sincerely, and I hope it will be analyzed.

Thank you, for now!

At. Ricardo A.

### DIFF
--- a/samples/MediatR.Examples.PublishStrategies/AsyncPingedHandler.cs
+++ b/samples/MediatR.Examples.PublishStrategies/AsyncPingedHandler.cs
@@ -13,7 +13,7 @@ namespace MediatR.Examples.PublishStrategies
 
         public string Name { get; set; }
 
-        public async Task Handle(Pinged notification, CancellationToken cancellationToken)
+        public async Task HandleAsync(Pinged notification, CancellationToken cancellationToken)
         {
             if (Name == "2")
             {
--- a/samples/MediatR.Examples.PublishStrategies/CustomMediator.cs
+++ b/samples/MediatR.Examples.PublishStrategies/CustomMediator.cs
@@ -14,7 +14,7 @@ namespace MediatR.Examples.PublishStrategies
             _publish = publish;
         }
 
-        protected override Task PublishCore(IEnumerable<Func<INotification, CancellationToken, Task>> allHandlers, INotification notification, CancellationToken cancellationToken)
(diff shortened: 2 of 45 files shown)


terrajobst:
I agree with your design premise that async methods should be suffixed with `Async`.

At the same time, I don't think it's useful to ask widely used libraries to make a big API breaking change. This is almost guaranteed to cause a lot of pain.

lilasquared:
I don't think it makes sense when every method is Async by default.  As the linked reference states, the point of the suffix was so that the function names didn't clash when Async methods were added in a nonbreaking version, or if the library supports both.  which MediatR does not.  I doubt this will ever happen for this library.

souzara:
@terrajobst  Maybe this PR, should be merged witha major version

souzara:
@lilasquared you're right, but it's not about technology, it's about the real cases, developers does not see if return is "Task", when async suffix is not present, is very commom that developer forget to use he the await keyword when call the asychronous method, I think that it is a "It's my precious" about what is right or wrong.



Reflect about this image:

UX


Maybe UX should be applied everywhere? :O

rellis-of-rhindleton:
It can be annoying when we make that mistake, @souzara, but when you’re using an IDE that offers Intellisense or the equivalent — which most C# devs are — you usually pick up on it quickly, and move on.

In the end this is a matter on which developers don’t all agree on a single approach. The library is maintained by devs that prefer not to have to deal with Async suffixes when there is no risk of confusing asynchronous and synchronous methods. (I agree with them, FWIW.) So you’re going to have a hard time selling the idea.

no1melman:
> async suffix help a lot all developers to avoid some simple bugs

I think both VS and Rider warn you that you're not `await`ing an async function.

Also, the async suffix is just something we got used to because we came from a sync world (with manual thread joining), and we should be getting used to a world where it is async all the way down as standard and sync is non-standard.

Any how, the function signature has the async information in it, probably doesn't need *more* documentation.

jbogard:
This discussion hasn't added anything new since the last discussions. I might decide differently today but it is not worth breaking the API after so many years at this point.

PR #561, opened 2020-09-23
show SnapshotTesting

SimonCropp:
would you consider using snapshot testing instead of traditional asserts?

uses https://github.com/VerifyTests/Verify

i have only converted over a subset to illustrate the approach


test/MediatR.Tests/ExceptionTests.Should_throw_argument_exception_for_publish_when_request_is_not_notification.verified.txt
```diff
@@ -0,0 +1,4 @@
+{
+  Type: 'ArgumentException',
+  Message: 'notification does not implement $INotification'
```
SimonCropp, line 3: note the $ is a bug

PR #563, opened 2020-09-25
Split IMediator interface in IPublisher and ISender interface

KenBonny:
The reason for this change is the quick [twitter chat](https://twitter.com/bonny_ken/status/1309182617653305344) I had with Jimmy about this.

### DIFF
--- a/test/MediatR.Tests/PipelineTests.cs
+++ b/test/MediatR.Tests/PipelineTests.cs
@@ -1,12 +1,13 @@
-using System.Collections.Generic;
 using System.Threading;
-using System.Threading.Tasks;
-using Shouldly;
-using StructureMap;
-using Xunit;
 
 namespace MediatR.Tests
 {
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Shouldly;
+    using StructureMap;
+    using Xunit;
+
     public class PipelineTests
     {
         public class Ping : IRequest<Pong>
--- a/test/MediatR.Tests/PipelineTests.cs
+++ b/test/MediatR.Tests/PipelineTests.cs
@@ -38,11 +38,10 @@ namespace MediatR.Tests
             {
                 _output = output;
             }
-
             public Task<Pong> Handle(Ping request, CancellationToken cancellationToken)
             {
                 _output.Messages.Add("Handler");
-                return Task.FromResult(new Pong {Message = request.Message + " Pong"});
+                return Task.FromResult(new Pong { Message = request.Message + " Pong" });
             }
         }
 
@@ -54,11 +53,10 @@ namespace MediatR.Tests
             {
                 _output = output;
             }
-
             public Task<Zong> Handle(Zing request, CancellationToken cancellationToken)
             {
                 _output.Messages.Add("Handler");
-                return Task.FromResult(new Zong {Message = request.Message + " Zong"});
+                return Task.FromResult(new Zong { Message = request.Message + " Zong" });
             }
         }
 
@@ -71,9 +69,7 @@ namespace MediatR.Tests
                 _output = output;
             }
 
-            public async Task<Pong> Handle(Ping request,
-                                           CancellationToken cancellationToken,
-                                           RequestHandlerDelegate<Pong> next)
+            public async Task<Pong> Handle(Ping request, CancellationToken cancellationToken, RequestHandlerDelegate<Pong> next)
             {
                 _output.Messages.Add("Outer before");
                 var response = await next();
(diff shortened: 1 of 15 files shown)


KenBonny:
Oh yes, I've removed unused usings as this is a pet peeve of mine. 😃

test/MediatR.Tests/PipelineTests.cs
```diff
@@ -1,14 +1,12 @@
...
 
 namespace MediatR.Tests
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Threading.Tasks;
-    using Shouldly;
-    using StructureMap;
-    using Xunit;
-
```
lilasquared, line 11: I understand removing unused references, but I don't know the implications of moving usings from inside the namespace to outside, any reason for doing so?

lilasquared:
@KenBonny once this is merged the dependency injection package should be updated as well to register the publisher and sender. https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection

test/MediatR.Tests/PipelineTests.cs
KenBonny: There is no benefit to using a `using` statement inside or outside of the namespace. I think the only difference is the scope. So the `using`s inside the namespace are only accessible inside the namespace and the `using`s outside are for the whole file. Since there is only one namespace in this file, it won't matter.

When I saw this in the file, my neurotic nature wanted to put them together as they belong together.

KenBonny: If you want, I can go over the code and put them all together and order them alphabetically.

lilasquared: If they are functionally the same it doesn't matter to me, I'll defer to Jimmy.  I've always put them outside but have seen them inside namespaces as well.

KenBonny:
@lilasquared Thanks for the heads up. I think the only thing that needs to happen is for the MediatR package to be updated. If the `IMediator` interface is registered, then I do believe most functionality should work. I did see that there is a line that registers all types in an assembly instead of registering everything by hand. I think that takes care of registering every type and interface in the assembly.

test/MediatR.Tests/PipelineTests.cs
KenBonny: Ok, I'll leave it up to @jbogard to decide if I roll this back, if I format and order the rest of the `using`s in the codebase or if this is ok.

lilasquared:
I'm pretty sure that only scans and searches for implementations of the handlers.  I'm referring to https://github.com/jbogard/MediatR.Extensions.Microsoft.DependencyInjection/blob/master/src/MediatR.Extensions.Microsoft.DependencyInjection/Registration/ServiceRegistrar.cs#L220
where the `IMediator` type is registered explicitly.  If you test your branch with that package and you try to inject `ISender` it fails to resolve.

test/MediatR.Tests/PipelineTests.cs
jbogard: I prefer consistency tbqh

KenBonny: Would you like me to roll it back or would you like me to look through the codebase and format the other places?

jbogard: If it's not consistent in the codebase today, leave it alone.

KenBonny: Should be rolled back now.

test/MediatR.Tests/PipelineTests.cs
```diff
@@ -2,7 +2,6 @@
```
no1melman, line 2: I kinda feel that this whole file should be reverted, bar the removing of the namespace - just inconsistent formatting

KenBonny: I rolled it back. On save, my IDE (Rider) formats all open files to my coding standards.

jbogard: So basically I need an `.editorconfig`

jbogard (approved):

KenBonny:
Is this ok, can it be merged or do other steps need to happen? That the other reviewers approve, for example. Or do I need to press a button to merge the PR?

KenBonny:
Don't mistake my question for impatience, but when will this be available in the package?

I'm asking so I can also update the DI package @lilasquared mentioned. I have a leaky memory (I should get that looked after) and if this takes a month or 2 (which is fine 😄 ), then I need to add a calendar notification to remind me to do the DI update then.

jbogard:
Tagged! Should be on NuGet shortly.

davidroth:
@jbogard  If you make such changes, please consider increasing the major version instead of the minor version in the future  ;-)

This interface split is technically a breaking change.
For example if an assembly "A" references IMediator  from 8.0.2, but at runtime 8.2. is supplied (assembly unification at build time), the interface split leads to the following runtime exception in "A" if it has not been recompiled with 8.2. 

```csharp
System.MissingMethodException : Method not found: 'System.Threading.Tasks.Task`1<!!0> MediatR.IMediator.Send(MediatR.IRequest`1<!!0>, System.Threading.CancellationToken)'.
```

Thanks 😃

PR #571, opened 2020-10-29
Update Simple Injector example

RyanMarcotte:
Updates the Simple Injector example to include `ISender` and `IPublisher` registrations


RyanMarcotte:
Per the [interface split PR](https://github.com/jbogard/MediatR/pull/563)), from an architecture perspective it may make more sense to register `ISender` and `IPublisher` only and not `IMediator`.  Thoughts?

Issue #458, opened 2019-11-15
Provide a RequestHandler interface that returns IAsyncEnumerable


adnang:
Now that we have C# 8.0, it would be nice if we had an interface such as `IStreamRequestHandler<TRequest, TResponse>` with method 
```
public async IAsyncEnumerable<TResponse> Handle(TRequest, CancellationToken)
{
}
```
Instead of needing another allocation for a `Task<IAsyncEnumerble<TResponse>>` using the existing `IRequestHandler<TRequest, IAsyncEnumerable<TResponse>>`.

lilasquared:
Is this really something that should be part of the base library?  It would be quite simple to have a seperate library `MediatR.AsyncEnumerable` or something of that nature with these types of simplified interfaces.

Shumigaj:
IAsyncEnumerable<T> interface is part of .NET Standard 2.1. If library migrates to 2.1 version it won't be possible to use it in the .NET Framework projects [.NET implementation support](https://docs.microsoft.com/en-us/dotnet/standard/net-standard)

wazzamatazz:
> 
> 
> IAsyncEnumerable interface is part of .NET Standard 2.1. If library migrates to 2.1 version it won't be possible to use it in the .NET Framework projects (read more [.NET implementation support](https://docs.microsoft.com/en-us/dotnet/standard/net-standard))

`IAsyncEnumerable` is available via the [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) package for projects targeting .NET Framework or .NET Standard 2.0. MediatR wouldn't need to use C# 8 language features to be able to add a request handler that returned `IAsyncEnumerable<T>`. 

Alternatively, it would be possible to include the idea proposed by @adnang using multi-targeting and only including it when targeting .NET Standard 2.1.

ThomasSkyldahl:
how should this be implemented if your are interested in a pull request?

I'm thinking of doing a Stream() method instead of Send() on the Mediator to handle the IAsyncEnumerable requests

Would that be a viable route?

jbogard:
Another route would be to only target `netstandard2.1`, that would be OK with me too, especially given the `net5.0` direction.

ThomasSkyldahl:
How would you handle the difference in signature with the single Send 

[...]

PR #574, opened 2020-11-04
Support for IAsyncEnumerable<TResponse> usage as CreateStream

driekus77:
Finally got some time to work on MediatR support for IAsyncEnumerable<TResponse>.

**Why IAsyncEnumerable?**
See issue https://github.com/jbogard/MediatR/issues/458
Personally I would like to use it in one of my new upcoming .NET Core projects.

**This Pull request is a first draft and needs:**
- Agreement on the Idea of supporting IAsyncEnumerable ;-)
- Agreement on naming of StreamRequestHandler interface/class and CreateStream Mediator method.
- Agreement on NETSTANDARD2_1/NETCOREAPP3_1 only (See project files and code)
- Agreement on ProjectReference mapping from UnitTest project to MediatR library its  TargetFrameworks:
   - net461  ==> net461
   - netcoreapp2.1 ==> netstandard2.0
   - netcoreapp3.1 ==> netstandard2.1

Review and/or Feedback comments are welcome!

(Unit)Tested it on Windows 10 | Visual Studio 2019: All OK


src/MediatR/IStreamRequestHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+    public interface IStreamRequestHandler<in TRequest, out TResponse>
+        where TRequest : IRequest<TResponse>
+    {
+        /// <summary>
+        /// Handles a stream request with IAsyncEnumerable as return type.
+        /// </summary>
+        /// <param name="request">The request</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>Response from the request</returns>
+        IAsyncEnumerable<TResponse> HandleAsync(TRequest request, CancellationToken cancellationToken);
```
lilasquared, line 22: For consistency with the other handler interfaces consider using just Handle rather than HandleAsync

src/MediatR/MediatR.csproj
```diff
@@ -1,39 +1,46 @@
...
+
+  <ItemGroup>
+    <None Include="..\..\assets\logo\gradient_128x128.png" Pack="true" PackagePath="" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+    <PackageReference Include="MinVer" Version="2.3.1" PrivateAssets="All" />
+  </ItemGroup>
+</Project>
```
lilasquared, line 46: did anything change in this file? Maybe line endings are wrong?

src/MediatR/IStreamRequestHandler.cs
driekus77: Renamed all HandleAsync to Handle.
Thanks for feedback!

src/MediatR/MediatR.csproj
driekus77: Reverted the line endings to master version of this file and reapplied my changes.

lilasquared:
General question, it seems that this does not plug in to the pipelines at all, do you have thoughts on if that should work with this as well some how?

driekus77:
I explicitly skipped the pipelines part because of:
- No experience with how pipelines and behaviours should work.
- The IAsyncEnumerable / Stream is already a Pipeline in itself and don't know what behaviour to tweak. Could become a whole lifecycle management on its own. 

I am open for any suggestions.

lilasquared:
Yeah.. that makes sense.  I've not dealt with the IAsyncEnumerable much so I'm not sure what the expected behavior even would be.  Given that pipelines are a huge extension point for this library I could see it being a problem if there's not some pattern for it.  One simple example is request validation which doesn't really care about the return type at all.  I would expect something like that to still be viable with this somehow.

lilasquared:
@driekus77 I have some code for the pipelines, would it be easier for me to fork your repo and create a pr into your branch or are you ok granting me permissions on your fork?

driekus77:
@lilasquared Thanks man: Looks great! I will give more detailed feedback in a few hours.

driekus77:
**Looked at @lilasquared his changes and not much to add.** 

Just a few remarks for StreamPipelineTests:
- Consistent with existing PipelineTests: Nice!
- StreamHandlerDelegate for next looks ok and again consistent with the rest. But the stream itself could also be used... (Maybe?)
- Functionality like stop the stream from propagating, validating (Not even starting), ..? all seems doable and up to the handler and/or receivers implementation.

**And a question: What about ReactiveX (Rx) its IObservable<TResponse>?**
We are now busy for supporting IAsyncEnumerable as a Pull-based Streaming solution. But other problems may need a Push based streaming solution like Rx its IObserable.
See for good comparison the Answer at SO: https://stackoverflow.com/questions/58471593/how-async-streams-compares-to-reactive-extension

**Pro's supporting a Push-Based Streaming solution like Rx:**
- Cover both flavors Pull and Push.
- Rx has much more to offer: Filtering, Combining etc... (Operations) , Events...

**Con's for Rx:**
- Probably need to include other packages....
- According to the Stackoverflow answer IAsyncEnumerable can act as source for a IObserable...

Disclaimer: I Only have good experience with the RxJs lib and not with Rx.Net (https://github.com/dotnet/reactive) 
This RxJs takes, at least for me, Stream handling to another level. Filtering, Mapping, Joining/Combining Streams and more.

Thanks again for your time and help!

lilasquared:
Given that almost all of the code for the new pipelines and handlers is new (and excluded unless you're using netstandard2.1 I am wondering if it would make more sense to bundle this as a seperate "extension" of mediator.  For it to be an extension some of the internal handler base classes could be exposed as well as the mediator properties and such as protected.  If we did that it could be much easier to start supporting the Rx IObservable since we could include the dependencies.  

With all that in mind though I am wondering what the gain is with this added functionality.  If streams and observables are already for the most part decoupled message handlers and can be extended through filtering, mapping, joining, etc. What is the value of this code?

If we do continue down this route still left todo is implementing all the built in pipeline  behaviors using the stream interface.

driekus77:
Rx IObservable was just me thinking out loud ;-)

I already tried to create the stream functionality as a separate extension but got stuck some how. If you wish I can try this again with current code? (MediatR.Streaming Project/lib for naming?)

MediatR and alike is excellent for me doing Command & Query Requests to "Open" up a single object response, Channel, Stream,.....  But leave the handling and receiving part up to its implementer with the restriction:  The wrapper(s) Task<TResponse> and now IAsyncEnumerable<TResponse> for Streams.
I already worked with IQueryHandler solution found at https://blogs.cuttingedge.it/steven/posts/2011/meanwhile-on-the-query-side-of-my-architecture/ in combination with IAsyncEnumerable and for me it worked perfect. The CQRS mechanism to open / pave the way and the Ad-Hoc streaming to give me the results. 
But as MediatR becomes a sort of standard in C# for CQRS I do prefer this added functionality inside MediatR.

I must admit: Overlooked the Build-in behaviors: WIll look into it now.

driekus77:
I See your point: The build-in behaviors need work. I'm now creating a subfolder/namespace MediatR.Pipeline.Streams with a Stream copy.

I'm bumping against a question/issue: Pre and Post processor on IAsyncEnumerable<TResponse> what to return: 
- For all TResponse Pre first then Handle then Post?  Pre Pre ... Handle Handle ... Post Post ...
- Or Per TResponse: Pre Handle Post Pre Handle Post Pre Handle Post ...

What todo with the Examples? Must they include the Streams as well? And if so how to implement this? Separate projects?

lilasquared:
Good questions.  I think initially replicating the existing behavior (Pre -> Handle -> Post) is sufficient.  This way it will give the ability to skip executing the async enumerable at all if necessary.  Because the async enumerables yield multiple results perhape another built in behavior could be `YieldPreProcessorBehavior` or the like which executes a set of pre-processors before yielding the result?  though I don't know the use case of such a behavior at the moment.  Sticking as close to the existing functionality will at least help us cover the scenarios already supported, and then we can move into more capabilities made possible by the streams.

lilasquared:
As for the samples at minimum I would add the stream interfaces into the existing samples (maybe using #if directive?).  Another enhancement will be to update the MediatR DI project to automatically include and scan for the new interfaces.

KuraiAndras:
Why not inlcude the new stream parts of the ISender for .NET Standard 2.0? The project can just have a dependency for [Microsoft.Bcl.AsyncInterfaces](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/) on .NET Standard 2.0.

driekus77:
@KuraiAndras I would like to get current setup working first. After that maybe move the code to a separate library.

driekus77:
Oke have a first draft setup with Pipeline.Stream but far from ready.

Bump into not being allowed to try/catch around a yield return.. 
See: https://stackoverflow.com/questions/346365/why-cant-yield-return-appear-inside-a-try-block-with-a-catch
This seems to be needed for the Exception handling to work with the Pre & Post Processors.
Don't have a workaround yet and help is welcome.

src/MediatR/Pipeline/Streams/StreamRequestPostProcessorBehavior.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            await foreach (var response in next().WithCancellation(cancellationToken).ConfigureAwait(false))
+            {
+                foreach (var processor in _postProcessors)
+                { 
+                    await processor.Process(request, response, cancellationToken).ConfigureAwait(false);
+
+                    yield return response;
+                }
+            }
```
lilasquared, line 33: This should be just the opposite execution of the PreProcessorBehavior, not nested foreach

driekus77:
Added the Pipeline.Streams UnitTests and fixed accordingly.
StreamRequestExceptionActionProcessorTests is still dummy because I don't know what functionality to expect.
Again Feedback is welcome.

driekus77:
Now have a Streaming Sample for SimpleInjector working. If this is the right way to go then, when I manage to get time for it, I will also do the rest of the samples.

Note that the MediatR.Example project now targets both "netstandard2.0"  and "netstandard2.1". This is needed for IAsyncEnumerable support.

Kind regards,

driekus77:
@lilasquared Am I on the right track with this PR or does it need a (big) refactor/fixing?

src/MediatR/Pipeline/Streams/StreamRequestExceptionActionProcessorBehavior.cs
```diff
@@ -0,0 +1,64 @@
...
+    public class StreamRequestExceptionActionProcessorBehavior<TRequest, TResponse> : IStreamPipelineBehavior<TRequest, TResponse>
+        where TRequest : notnull
+    {
+        private readonly ServiceFactory _serviceFactory;
+
+        public StreamRequestExceptionActionProcessorBehavior(ServiceFactory serviceFactory) => _serviceFactory = serviceFactory;
+
+        public async IAsyncEnumerable<TResponse> Handle(TRequest request, [EnumeratorCancellation] CancellationToken cancellationToken, StreamHandlerDelegate<TResponse> next)
+        {
+            var asyncEnum = next().WithCancellation(cancellationToken).ConfigureAwait(false).GetAsyncEnumerator();
```
Tornhoof, line 28: If you do the asyncenum dance youreself via movenextasync, aren't you supposed to dispose it yourself too?
The same applies to the behaviour below too.

src/MediatR/Pipeline/Streams/StreamRequestExceptionProcessorBehavior.cs
```diff
@@ -0,0 +1,91 @@
...
+        private readonly ServiceFactory _serviceFactory;
+
+        public StreamRequestExceptionProcessorBehavior(ServiceFactory serviceFactory) => _serviceFactory = serviceFactory;
+
+        public async IAsyncEnumerable<TResponse> Handle(TRequest request, [EnumeratorCancellation] CancellationToken cancellationToken, StreamHandlerDelegate<TResponse> next)
+        {
+            var asyncEnum = next().WithCancellation(cancellationToken).ConfigureAwait(false).GetAsyncEnumerator();
+
+            bool thrown = false;
+            StreamRequestExceptionHandlerState<TResponse> state = new StreamRequestExceptionHandlerState<TResponse>();
```
Tornhoof, line 31: maybe do the ``new()`` only on demand if an exception is thrown?

gunndabad:
Keen to see support for `IAsyncEnumerable` here but I'm not crazy on the 'Stream' part of the naming. In my mind something named 'Stream' would deal with `Stream`s. That said, I can't think of a better alternative right now 😊

src/MediatR/Internal/StreamRequestHandlerWrapper.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public override IAsyncEnumerable<TResponse> Handle(IRequest<TResponse> request, CancellationToken cancellationToken, ServiceFactory serviceFactory)
+        {
+            IAsyncEnumerable<TResponse> Handler() => GetHandler<IStreamRequestHandler<TRequest, TResponse>>(serviceFactory).Handle((TRequest) request, cancellationToken);
+
+            return serviceFactory
+                .GetInstances<IStreamPipelineBehavior<TRequest, TResponse>>()
+                .Reverse()
+                .Aggregate((StreamHandlerDelegate<TResponse>) Handler, (next, pipeline) => () => pipeline.Handle((TRequest) request, cancellationToken, next))();
```
gregory-seidman, line 38: Can the `.WithCancellation(cancellationToken).ConfigureAwait(false)` (as in `GenericStreamPipelineBehavior.cs` in the stream example) be done here instead of expecting the implementation to handle it? Specifically, replace
```csharp
pipeline.Handle((TRequest) request, cancellationToken, next)
```
with
```csharp
pipeline.Handle((TRequest) request, cancellationToken, () => next().WithCancellation(cancellationToken).ConfigureAwait(false))
```

driekus77: Thanks for the review comments. Made a change in the code. Please review again because of compile errors I had to alter your suggested change a little bit.

driekus77:
> Keen to see support for `IAsyncEnumerable` here but I'm not crazy on the 'Stream' part of the naming. In my mind something named 'Stream' would deal with `Stream`s. That said, I can't think of a better alternative right now 😊

Thanks for the review comment. I'm with you with the naming but could find a better name for it. Alternative proposals are welcome!. (OpenFlow, OpenAsyncEnumberable, ...?)

src/MediatR/Pipeline/Streams/StreamRequestExceptionActionProcessorBehavior.cs
driekus77: Thanks for the review comment. I will dive into it.

src/MediatR/Pipeline/Streams/StreamRequestExceptionProcessorBehavior.cs
driekus77: Good catch. Fixed it. Thanks.

driekus77:
Hi All,

Sorry for my delay in picking up the review comments simply didn't had time for it.

In order to get this PR implemented I will need some help from people who have experience with the pipeline behaviors. The basic things are now working but the Tests and Samples need more work.

Thanks for the review comments and help.

src/MediatR/Internal/StreamRequestHandlerWrapper.cs
gregory-seidman: I did a little research and found [this instructive blog post](https://oleh-zheleznyak.blogspot.com/2020/07/enumeratorcancellation.html). I see two options:
1. Put the `EnumeratorCancellation` attribute on the cancellation token argument in the implementations of the stream handler (examples and `StreamRequestExceptionActionProcessorBehavior` and such, plus you're already doing it in the tests) and hope everyone using the library follows that practice (which they should).
2. Wrap `next()` as `() => NextWrapper(next(), cancellationToken)` with the following implementation:
```csharp
private static IAsyncEnumerable<T> NextWrapper<T>(
    IAsyncEnumerable<T> items,
    [EnumeratorCancellation] CancellationToken cancellation)
{
    var cancellable = items
        .WithCancellation(cancellationToken)
        .ConfigureAwait(false);
    await foreach (var i in cancellable)
    {
        yield return i;
    }
}
```

In either case, you should remove the `.WithCancellation(cancellationToken).ConfigureAwait(false)` from everywhere else.

driekus77:
Hi All,

Took some time to look closer at the unit tests, while fixing review comments from @gregory-seidman. 
Discovered that the PreProcess only is processed once at the beginning of the stream and the PostProcess after every item in the IAsyncEnumerable.

I tried to emit the preprocess for every item but this doesn't seem doable. Because you don't know upfront if an item is coming or not. Only thing I can think of right now is wrapping the whole next() delegate in another delegate and call all Pipeline and handlers from their. But this means a big refactoring for StreamRequestHandlerWrapper...

Help is welcome. Personally I don't use the Pipelines and behaviors that much.

jbogard:
It would be weird to me to have pipelines/behaviors run before/after each item. That seems better suited for something like System.Linq.Async, no?

driekus77:
Thanks for quick reply. 

Yes I totally agree on this. 
But which valid use cases are then left for pipelines and behaviors on IAsyncEnumerable stream(s).
Can a library like System.Linq.Async or maybe Rx.Net (https://github.com/dotnet/reactive) handle all the use case better?

jbogard:
To me, pipelines are about cross-cutting concerns about the overall request, not items in the response. Items in the response are the responsibility of the handler.

jbogard:
So basically - you can have a pipeline, but I would leave out the more granular ones (pre/post processor, exceptions) since those were very specific to a request/response

driekus77:
Oke thats clearer now. 

My assumption for this PR would be:
- Leave exceptionhandling for IAsyncEnumerable out
- No PreProcessor(s)
- No PostProcessor(s)
- Extend some unit testing with multiple streaming... (If I have time left)
- Am I forgetting something?...

jbogard:
BTW the reason I'm finally looking at this is I'm working with gRPC streams 😆

driekus77:
Oke removed a lot of files regarding exception handling, Pre- and, Post- Processing for IAsyncEnumerable. 
The sample for GenericStreamPipelineBehavior is still working and added an Unit Test for it.

driekus77:
Dear all, 

**Started over again with following changes:**
- Leave MediatR Lean and Mean with only some api changes that are needed in order to get e.g. AsyncEnumerable working.
- --> Make ServiceFactory available as property
- --> Make HandlerBase public
- AsyncEnumerable now working through Extensions in its own library. 
- --> See  https://github.com/driekus77/MediatR/blob/feature/extendable/src/MediatR.AsyncEnumerable/MediatorExtensions.cs
- Naming change (Test)
- --> All Stream naming renamed to AsyncEnumerable for clarity 
- --> CreateStream in MediatorExtensions renamed to "ToAsyncEnumerable(..."

**Which code change proposal tastes beter?** 
--> This PR with all of Stream inside the MediatR API?
--> The "feature/extendable" branch with AsyncEnumerable naming and most related code in dedicated library?
--> A mix?
--> ....?

**Why?**
Above Feedback in comments state that this PR code change is big. Also naming like "CreateStream"  was not appreciated by everybody. So in order to give people some choice an extra implementation.

Keen to hear some feedback.

Thanks and kind regards

driekus77:
Fixed issues and added some features in the alternative implementation "feature/extendable":
   https://github.com/driekus77/MediatR/tree/feature/extendable

- Added IAsyncEnumerablePipeline tests
- Fixed building extra nuget package (Build.ps1)
- Signed MediatR.AsyncEnumerable with same key as MediatR
- Added System.Linq.Async package to MediatR.AsyncEnumerable.Tests Unit Tests lib.
- Moved nullable support to project i.s.o. per file
- Added netstandard2.0 support with **Microsoft.Bcl.AsyncInterfaces** reference to MediatR.AsyncEnumerable library

@lilasquared, @gregory-seidman, @Tornhoof , @jbogard: Which implementation do you prefer? 

Should I create a new PR from alternative branch or leave this PR as is?

gregory-seidman:
> Fixed issues and added some features in the alternative implementation "feature/extendable":
> https://github.com/driekus77/MediatR/tree/feature/extendable
> 
> * Added IAsyncEnumerablePipeline tests
> * Fixed building extra nuget package (Build.ps1)
> * Signed MediatR.AsyncEnumerable with same key as MediatR
> * Added System.Linq.Async package to MediatR.AsyncEnumerable.Tests Unit Tests lib.
> * Moved nullable support to project i.s.o. per file
> * Added netstandard2.0 support with **Microsoft.Bcl.AsyncInterfaces** reference to MediatR.AsyncEnumerable library
> 
> @lilasquared, @gregory-seidman, @Tornhoof , @jbogard: Which implementation do you prefer?
> 
> Should I create a new PR from alternative branch or leave this PR as is?

Can you make a (draft) PR with your new branch to make it easier to see what's going on?

NeussConsulting:
Please merge this PR

brian-pickens:
Also keen on an IAsyncEnumerable implementation. Where are we at with the PR?

jbogard:
@driekus77 yes can you create a draft PR?

jbogard (approved):

jbogard:
FYI I created a PR to separate the stream request into its own interface to avoid mistakes here #685

PR #575, opened 2020-11-05
Using C# 9.0 features

jbogard:


src/MediatR/Unit.cs
```diff
@@ -29,10 +29,7 @@ public struct Unit : IEquatable<Unit>, IComparable<Unit>, IComparable
         ///  - Zero: This object is equal to <paramref name="other" />.
         ///  - Greater than zero: This object is greater than <paramref name="other" />.
         /// </returns>
-        public int CompareTo(Unit other)
-        {
-            return 0;
-        }
+        public int CompareTo(Unit other) => 0;
```
lilasquared, line 32: Is there a change to editorconfig you can do to make this preferred?

jbogard: Probably, but it'd be R#-only

src/MediatR/Internal/RequestHandlerWrapper.cs
```diff
@@ -48,7 +48,7 @@ internal class RequestHandlerWrapperImpl<TRequest, TResponse> : RequestHandlerWr
             return Handle((IRequest<TResponse>)request, cancellationToken, serviceFactory)
                 .ContinueWith(t =>
                 {
-                    if (t.IsFaulted)
+                    if (t.IsFaulted && t.Exception?.InnerException is not null)
```
no1melman, line 51: Would this break functionality - as in, does `ExceptionDispatchInfo` still throw even if the innerEx is null?

jbogard: Yes it does, https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.capture?view=netcore-3.1#exceptions

PR #580, opened 2020-11-10
Make wrappers public, move them to new namespace

### LINKED ISSUE #572: Make RequestHandlerBase and related classes public
I'm using MediatR with Unity 3D, with the IL2CPP backend, and found a shortcoming in IL2CPP which will make MediatR unusable for IRequests which return value types. [Unity forum thread](https://forum.unity.com/threads/il2cpp-does-not-generate-constructor-for-generic-classes-that-have-struct-as-parameter.997274/#post-6476099), [Repro](https://github.com/KuraiAndras/IL2CPPBugSample).

The problem boiled down is that IL2CPP will not generate an implementation for RequestHandlerWrapperImpl in cases where the generic arguments contain a value type. [Relevant Unity blog post](https://blogs.unity3d.com/2015/06/16/il2cpp-internals-generic-sharing-implementation/)

A workaround for this as suggested by @VolodymyrBS is to manually create instances of RequestHandlerWrapperImpl with the problematic generic type parameters. For this to work you either have to fork MediatR, make the classes in question public and keep the fork in sync with the base repo, or use something like [Open Sesame Compiler](https://github.com/mob-sakai/OpenSesame) which will let you access internal classes, or refactor the code so that all requests return reference types (and introduce some Unit equivalent which is a reference type). The Open Sesame solution although works, and provides the least amount of maintenance work for an end user, it will force you to use the compiler, and suddenly expose every internal/private class. The nicest solution for this problem would be if the RequestHandlerBase and other related classes would become public.

This could be problematic:
- People might assume that they want to use these classes even though it does not provide benefit for them (This might be solved with clear documentation)
- Might break libraries if they have similarly named classes (I suppose this is really an edge case)
- Might break libraries where these classes are used in reflection and checked whether the class has an internal modifier (A really slim edge case IMO)

What do you think about this? Are you comfortable with exposing these internals in the library, or should I just use the Open Sesame compiler? If you say yes, I will create the PR.

Also, in the linked forum thread I mentioned creating a Unity Editor utility which will generate the needed stub calls while in the Unity Editor and I will make it FOSS so that people using Unity and MediatR won't have to deal with this problem.

KuraiAndras:
Closes #572

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -1,6 +1,6 @@
 namespace MediatR
 {
-    using Internal;
+    using Wrappers;
     using System;
     using System.Collections.Concurrent;
     using System.Collections.Generic;
--- a/src/MediatR/Wrappers/NotificationHandlerWrapper.cs
+++ b/src/MediatR/Wrappers/NotificationHandlerWrapper.cs
@@ -1,4 +1,4 @@
-namespace MediatR.Internal
+namespace MediatR.Wrappers
 {
     using System;
     using System.Collections.Generic;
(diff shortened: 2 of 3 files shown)


KuraiAndras:
Should I add documentation for the moved types?

jbogard:
Nah.

jbogard:
Conflicts :(

KuraiAndras:
I'll get to it soon.

KuraiAndras:
@jbogard  This should be good now

jbogard (approved):

KuraiAndras:
🎉

PR #623, opened 2021-03-15
feat: allow handlers to return a 'null' response

kevinbuhmann:
I have a use case where I want response handlers to be able to return null. Using `return null!` works, so there is nothing in the library preventing null responses except the nullable metadata. I know changing the nullable metadata is breaking change, but I thought I would open this PR for consideration anyway so that the nullability metadata can be made accurate.

Sorry if there is an existing issue for this. I did search, but did not find one.


eskaufel:
I vote against this as it would give me a lot of work.

jjroman:
I think you could implement the NULL pattern in any response. The null pattern will help you to cover all the scenarios. NULL values are bad options most of the time.

jbogard:
Yes, I think you'd be better off implementing the Null/Maybe pattern in this case. I've never had a need to return `null` because I always expect *some* response.

PR #665, opened 2021-10-25
Allows for generic request handlers to be resolved.

WillMarcouiller:
For more information please refer to this so question: https://stackoverflow.com/questions/69546722/ninject-just-wont-register-mediatr-irequesthandler-using-convention-based-bi

### DIFF
--- a/samples/MediatR.Examples.Ninject/ContravariantBindingResolver.cs
+++ b/samples/MediatR.Examples.Ninject/ContravariantBindingResolver.cs
@@ -20,10 +20,9 @@
             {
                 var genericType = service.GetGenericTypeDefinition();
                 var genericArguments = genericType.GetGenericArguments();
-                if (genericArguments.Count() == 1
-                 && genericArguments.Single().GenericParameterAttributes.HasFlag(GenericParameterAttributes.Contravariant))
+                if (IsContravariant(genericArguments))
                 {
-                    var argument = service.GetGenericArguments().Single();
+                    var argument = genericArguments.Single();
                     var matches = bindings.Where(kvp => kvp.Key.IsGenericType
                                                            && kvp.Key.GetGenericTypeDefinition().Equals(genericType)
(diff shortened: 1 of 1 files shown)


jbogard:
Can you add a test to show a scenario that this addresses?

WillMarcouiller:
> Can you add a test to show a scenario that this addresses?

Just saw your answer, sorry for delay.  In fact, I just can't see what kind of a test I could write to demonstrate what it addresses apart the given SO link provided (https://stackoverflow.com/questions/69546722/ninject-just-wont-register-mediatr-irequesthandler-using-convention-based-bi).  

I'll try and perhaps come up with something.  In the meantime, I would not have created this request if I had not experienced any trouble.

jbogard:
No I understand you had an issue, but I don't like taking PRs that change behavior without a unit test that demonstrates what behavior this addresses. Otherwise I might break this behavior in the future.

WillMarcouiller:
> No I understand you had an issue, but I don't like taking PRs that change behavior without a unit test that demonstrates what behavior this addresses. Otherwise I might break this behavior in the future.

I totally understand it.  I'll write a test to cover this issue then.  :)

WillMarcouiller:
> No I understand you had an issue, but I don't like taking PRs that change behavior without a unit test that demonstrates what behavior this addresses. Otherwise I might break this behavior in the future.

I'm gonna need a hand to make the solution work on my computer as I just can't compile the cloned project.

First, I cloned using:

> gh repo clone jbogard/MediatR gh/MediatR
> cd /gh/MediatR
> gh pr checkout 665

Then I can see my changes within the ContravariantBindingResolver class in the Ninject sample.

What I don't understand is how can't I even make it restore and compile?  Here's a screenshot of the error list.  This happens when I rebuild the entire solution.  It seems to bug on Ninject, and I suspect the problem is deeper than that.  Help would be appreciated.  I'm not used to contribute on gh.

image

PR #670, opened 2021-11-08
Use explicit Interface types vs object in ISender and IPublisher.

StevenTCramer:
Update ISender Send method to take explicit `IRequest` instead of `object`
Update IPublisher Publish method to take explicit `INotification` instead of `object`

Giving us compile time checks vs runtime.

## Tests
Associated tests cases that are no longer required have been removed.
All other tests continue to pass.

I am not sure if this is a breaking change.  Technically the interface changes and thus some peoples code may not compile if they were using explicit `object`.  Which would mean this is a breaking change.  But they would have gotten a runtime error, where this will give a compile time error.

### DIFF
--- a/samples/MediatR.Examples.PublishStrategies/Publisher.cs
+++ b/samples/MediatR.Examples.PublishStrategies/Publisher.cs
@@ -25,22 +25,22 @@ namespace MediatR.Examples.PublishStrategies
         public IDictionary<PublishStrategy, IMediator> PublishStrategies = new Dictionary<PublishStrategy, IMediator>();
         public PublishStrategy DefaultStrategy { get; set; } = PublishStrategy.SyncContinueOnException;
 
-        public Task Publish<TNotification>(TNotification notification)
+        public Task Publish<TNotification>(TNotification notification) where TNotification : INotification
         {
             return Publish(notification, DefaultStrategy, default(CancellationToken));
         }
 
-        public Task Publish<TNotification>(TNotification notification, PublishStrategy strategy)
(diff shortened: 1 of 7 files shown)


lilasquared:
the object api was added explicitly in this PR https://github.com/jbogard/MediatR/pull/415/files why are you wanting to remove it?

jbogard:
See discussion on #385 too. There are two changes here too:
- Adding generic constraints
- Going from `object` to `IRequest`/`INotification`.

The `object` scenario supports cases where the types aren't known until runtime. Changing to `IRequest` won't result in a compile-time error, but a runtime exception:

```csharp
object requestObject = DeserializeRequest(incomingJson);
IRequest request = (IRequest) requestObject; // Runtime exception if it doesn't match
await mediator.Send(request);
```

The exception will be the same whether MediatR does the casting or the consumer does it. I'd rather keep `object` to make this API very explicit about the intent, that you don't know the type at compile-time, only at runtime.

StevenTCramer:
Ah I am reading #385 now and didn't think of this use case.  I vaguely remember reading something about this before.  I was updating my architecture and was wondering why this wasn't explicit. Did a quick search and couldn't find a reason ( guess I didn't look hard enough)

jbogard:
Ha no worries, I couldn't remember either and had to use blame to find out why. It was a good thought but would break existing folks for this scenario.

StevenTCramer:
Yep,  I'm just used to things that are not strongly typed being a smell.  And implemented a delegate for ISender in my tests to wrap everything in a scope and wondered... why is the one strong typed and the other not.  :)

Itch scratched now I know.  Thanks.

PR #671, opened 2021-11-09
Support for only netstandard2.1

jbogard:
With .NET 6 released, this PR changes the TFNs to only `netstandard2.1`, which then would solely target MediatR for .NET Core 3.0 and above. This would be 2 LTS versions of .NET supported:
- 3.1
- 6.0


remcoros:
So no more .NET Framework? Seems unnecessary to remove that, right?

jbogard:
@remcoros With the stream support, I don't want to do the whole conditional compilation thing

remcoros:
Understandable, easy enough to fork anyway ^-^

mantasaudickas:
Was there really a need to target netstandard2.1 in MediatR.Contracts?

jbogard:
Versus...?

mantasaudickas:
> Versus...?

Well.. now I need to upgrade a bunch of our contract projects to netstandard2.1 which is not really needed.
Imho MediatR.Contracts can stay at netstandard2.0 without any issues.

PR #673, opened 2021-11-17
Add support to enable generic based interface defined handlers

gjdodd:
When using Lamar on a project which made heavy use of generics, interfaces, and proxy objects this all worked fine. When it was used inside a project which used castle.windsor for the container, the handlers could not be loaded from the container as the use of the actual type of the request would not be registered in the container, but the underlying interfaces it was based on were.

This change allows for the send to support passing the request type to allow for a cleaner hint for the container to find the actual handler. 

I have included a tests which shows the exception of the container not able to find the handler, then one which uses the new method to load based on the interfaces

### DIFF
--- a/test/MediatR.Tests/GenericTypeConstraintsTests.cs
+++ b/test/MediatR.Tests/GenericTypeConstraintsTests.cs
@@ -101,8 +101,10 @@ namespace MediatR.Tests
                 });
                 cfg.For<ServiceFactory>().Use<ServiceFactory>(ctx => ctx.GetInstance);
                 cfg.For<IMediator>().Use<Mediator>();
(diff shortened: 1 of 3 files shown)


jbogard:
I avoided this overload because it's ugly - you have to specify all the generic arguments all the time

test/MediatR.Tests/GenericTypeConstraintsTests.cs
```diff
@@ -83,7 +83,7 @@ public Task<Pong> Handle(Ping request, CancellationToken cancellationToken)
             {
                 return Task.FromResult(new Pong { Message = request.Message + " Pong" });
             }
-        }
+        }
```
no1melman, line 86: can remove whitespace addition

PR #710, opened 2022-02-11
Bring back support for netstandard2.0 and net461

### LINKED ISSUE #691: .NET48 no longer supported
Why has this support been dropped? We have been using this with our .NET48 solutions for several years but as of version 10 it is no longer possible to update / install the package.

> Could not install package 'MediatR 10.0.1'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.8', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

GerardSmit:
.NET Standard 2.0 and .NET Framework 4.6.1 support was removed because `IAsyncEnumerable` isn't supported:

> You can continue to use v9 just fine. V10 includes support for `IAsyncEnumerable` which doesn't have much support in full .NET. Since I don't use .NET 4.x and haven't for many years, I decided to move on to the LTS releases of .NET Core.
> https://github.com/jbogard/MediatR/issues/691#issuecomment-1009966065

Microsoft created a NuGet packages that brings support to `IAsyncEnumerable` to .NET Standard 2.0 and  .NET Framework 4.6.1: https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/.
With this NuGet package, the latest commit in `master` [builds and all unit tests succeed on .NET Framework 4.6.1](https://github.com/GerardSmit/MediatR/runs/5159457252#step:5:190).

**Note:** I haven't reverted the `Microsoft.NETFramework.ReferenceAssemblies` reference that was removed in #671. The CI is running on Windows, so this package isn't necessarily. I can revert this back if needed.

---

Fixes #691
Reverts #671

### DIFF
--- a/src/MediatR/MediatR.csproj
+++ b/src/MediatR/MediatR.csproj
@@ -4,7 +4,7 @@
     <Authors>Jimmy Bogard</Authors>
     <Description>Simple, unambitious mediator implementation in .NET</Description>
     <Copyright>Copyright Jimmy Bogard</Copyright>
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>net461;netstandard2.0;netstandard2.1</TargetFrameworks>
     <Nullable>enable</Nullable>
     <Features>strict</Features>
     <PackageTags>mediator;request;response;queries;commands;notifications</PackageTags>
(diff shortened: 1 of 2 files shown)


jbogard:
This doesn't work on my primary workstation, M1 Macbook Pro.

Shortly I'm moving the build agents to Ubuntu as well. Sorry for the inconvenience!

PR #743, opened 2022-05-20
Small improvement of MediatR.Examples.Windsor

alexandruchirita4192:
Better checking of baseRequestType and checking assumptions with Debug.Assert instead of writing assumptions as comments


alexandruchirita4192:
Closing this pull request because I'll open another one with more important fixes.

PR #753, opened 2022-06-15
fix bug concurrentdictionary GetOrAdd

EngRajabi:
I had some big changes
1- As in the following documents. Using the GetOrAdd method is by no means Thread Safe. I added the test. If you test in Func during GetOrAdd, func will be called several times. But in the end one remains. This adds a lot of overhead to the program.
[microsoft](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-6.0)
[andrewlock](https://andrewlock.net/making-getoradd-on-concurrentdictionary-thread-safe-using-lazy/)
2- Add an analyzer to increase the quality and maintenance of the code

### DIFF
--- a/Clean.bat
+++ b/Clean.bat
@@ -1,0 +1,6 @@
+FOR /d /r . %%d in (bin,obj) do @if exist "%%d" rd /s/q "%%d"
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S Packages') DO RMDIR /S /Q "%%G"
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S .vs') DO RMDIR /S /Q "%%G"
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S TestResults') DO RMDIR /S /Q "%%G"
+FOR /F "tokens=*" %%G IN ('DIR /B /AD /S AppPackages') DO RMDIR /S /Q "%%G"
+DEL /Q /F /S "*.csproj.user"
--- a/Clean.bat
+++ b/Clean.bat
@@ -1,6 +1,0 @@
-FOR /d /r . %%d in (bin,obj) do @if exist "%%d" rd /s/q "%%d"
-FOR /F "tokens=*" %%G IN ('DIR /B /AD /S Packages') DO RMDIR /S /Q "%%G"
-FOR /F "tokens=*" %%G IN ('DIR /B /AD /S .vs') DO RMDIR /S /Q "%%G"
-FOR /F "tokens=*" %%G IN ('DIR /B /AD /S TestResults') DO RMDIR /S /Q "%%G"
-FOR /F "tokens=*" %%G IN ('DIR /B /AD /S AppPackages') DO RMDIR /S /Q "%%G"
-DEL /Q /F /S "*.csproj.user"
--- a/samples/MediatR.Examples.PublishStrategies/Publisher.cs
+++ b/samples/MediatR.Examples.PublishStrategies/Publisher.cs
@@ -78,7 +78,7 @@ public class Publisher
     {
         foreach (var handler in handlers)
         {
-            var task = Task.Run(() => handler(notification, cancellationToken));
+            _ = Task.Run(() => handler(notification, cancellationToken));
         }
 
         return Task.CompletedTask;
--- a/samples/MediatR.Examples.PublishStrategies/Publisher.cs
+++ b/samples/MediatR.Examples.PublishStrategies/Publisher.cs
@@ -78,7 +78,7 @@ public class Publisher
     {
         foreach (var handler in handlers)
         {
-            Task.Run(() => handler(notification, cancellationToken));
+            var task = Task.Run(() => handler(notification, cancellationToken));
         }
 
         return Task.CompletedTask;
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,21 @@
 <Project>
   <PropertyGroup>
     <LangVersion>10.0</LangVersion>
-    <NoWarn>$(NoWarn);CS1701;CS1702;CS1591</NoWarn>
+    <NoWarn>$(NoWarn);CS1701;CS1702;CS1591;VSTHRD200</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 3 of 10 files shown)


samples/MediatR.Examples.PublishStrategies/Publisher.cs
```diff
@@ -78,7 +78,7 @@ private Task ParallelNoWait(IEnumerable<Func<INotification, CancellationToken, T
     {
         foreach (var handler in handlers)
         {
-            Task.Run(() => handler(notification, cancellationToken));
+            var task = Task.Run(() => handler(notification, cancellationToken));
```
jbogard, line 81: Why do this?

poke:
It's not clear to me what the bug is that you're trying to fix here.

Yes, the creation callback of `GetOrAdd` is not guaranteed to run only once in a concurrent situation but if you're looking at the usages here, then this is only some object creation. And the objects that are created here, e.g. `RequestHandlerWrapper`, don't really have any particular overhead. They would probably be even cheaper than a `Lazy`.

So before going this route “just because”, I would prefer seeing a benchmark of this first to see what's the impact of the Lazy wrappers is compared to the overhead of possibly throwing away handler wrapper instances.

Clean.bat
```diff
@@ -0,0 +1,6 @@
+FOR /d /r . %%d in (bin,obj) do @if exist "%%d" rd /s/q "%%d"
```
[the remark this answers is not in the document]
EngRajabi: I deleted this file

samples/MediatR.Examples.PublishStrategies/Publisher.cs
EngRajabi: Thankful
Yes, I did because of that
https://github.com/Microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD110.md

peteraritchie:
A lot of noise in this PR.  Separating out style changes and warning fixes would go a long way to making the "fix" clear.

peteraritchie:
Also, does Should_Concurrency_MultipleInvoke_handler fail without the other changes?

remcoros:
This trick (ConcurrentDictionary with a Lazy wrapping the creation callback) is generally used when the creation is an expensive operation (like in asp.net, creating the middleware handler pipeline or querying a database).

As an example, take a simple blog app, using a ConcurrentDictionary as a cache for blog posts, and a creation callback that calls into a database.
The first request misses the key lookup and a database query will run. If a 2nd request comes in and the 1st query hasn't finished yet, it will also miss the key lookup on the dictionary and trigger another create callback (and with it, a second database query).
This could cause some redundant database queries. but normally not a problem. Under high load though, this could cause a domino effect on the database.
When using a ConcurrentDictionary of Lazy values, where Lazy is blocking. This will make it so the 2nd request will synchronously block until the database query (filling the cache) of request 1 is finished.
If that is desirable, ultimately depends on the use case in those specific situations.

for MediatR, the creation is simple and doesn't call any external 'slow' resources, so adding Lazy here would only hurt performance under high load because of lock contention.

EngRajabi:
I do not agree with this.  It is true that in certain scenarios, using no lazy will cause very, very serious damage to the program.  But in mediatr this cost is also high.  If building obj is not expensive, why use dic and not build it every time?  It is better to make that obj once.  Consider this.  In very large and high load projects, too many objs are built.  What's wrong with making lazy for the first time?  The point is that being lazy has no effect on get

poke:
> But in mediatr this cost is also high.

Do you have any numbers to back this statement up? In order to suggest such a change, you should be able to prove that the duplicate object creation is indeed a problem and that the implementation with Lazy performs better.

My expectation would be that Lazy could perform better in the extreme situations where a lot of requests hit the mediator initially (before the cache is available) but that it will perform worse in any other situation, i.e. both directly after the cache has been created (since then you will need to go through the Lazy for every single request), and for the way more likely average case where there are not a high number of initial requests.

The Lazy only helps during the time before the object creation has completed; a process that should be done in a few milliseconds. So it can only make the performance better for that very small window, while equally hurting the performance in every other case.

EngRajabi:
To evaluate the better performance, divide it into 2 parts.  
1.  Creating obj 
2. Obtaining obj
in section 2 it does not matter whether obj was created via lazy or without it, because it was created once and stored in dic.  In Section 1, when the load is high and the number of handlers is large, they make the whole obj unused

poke:
> it does not matter whether obj was created via lazy or without it

This incorrectly assumes that `Lazy<T>` is free which it is not, as can be shown with a simple benchmark utilizing the `LazyConcurrentDictionary` you are proposing here:

```csharp
public class Benchmarks
{
    private readonly ConcurrentDictionary<Type, IHandler> _handlers = new();
    private readonly LazyConcurrentDictionary<Type, IHandler> _handlersLazy = new();

    [GlobalSetup]
    public void Setup()
    {
        // ensure the handler is already created in both dictionaries
        _ = _handlers.GetOrAdd(typeof(MyHandler), _ => new MyHandler());
        _ = _handlersLazy.GetOrAdd(typeof(MyHandler), _ => new MyHandler());
    }

    [Benchmark]
    public IHandler Normal()
        => _handlers.GetOrAdd(typeof(MyHandler), _ => throw new NotSupportedException());

    [Benchmark]
    public IHandler Lazy()
        => _handlersLazy.GetOrAdd(typeof(MyHandler), _ => throw new NotSupportedException());
}

public class MyHandler : IHandler { }
public interface IHandler { }
```

```
| Method |     Mean |    Error |   StdDev |
|------- |---------:|---------:|---------:|
| Normal | 16.86 ns | 0.321 ns | 0.268 ns |
|   Lazy | 30.98 ns | 0.607 ns | 0.768 ns |
```

> In Section 1, when the load is high and the number of handlers is large

The number of handlers and the general load is irrelevant for this. What matters is the _initial load_ for a particular handler. I.e. the amount of concurrent requests that hit the mediator *before* the handler wrapper was created for the first time. And only for the very short window until the object was created and stored in the dictionary.

So assuming a normal application that stays alive for a longer duration, this “optimization” would only benefit a very particular and unrealistic situation for a short time, while equally affecting the whole remainder of the application’s lifetime.

Throwing away a few wrapper instances in such a special case is not as much of a problem as you want to make it.

PR #767, opened 2022-07-19
Each matching exception handler/action is invoked at most once

FDUdannychen:
Currently base exception handlers/actions will be executed multiple times when a derived exception is thrown. 
For example you have `MyExceptionHandler:IRequestExceptionHandler<MyRequest, MyResponse>`, it will be invoked twice when `MyException:Exception` is thrown. Its invocation count grows with the exception inheritance.

This PR fixes this bug by checking handler/action type, please note that checking interface types is not enough because different interfaces can share the same handler/action type.


inary-wrk:
@jbogard Can you please review this PR.

jbogard (approved):

Issue #692, opened 2022-01-11
Could not load type Mediatr.IRequest


bend:
Hi, 
Since update to Mediatr 10, I have the following error at runtime: 
` Could not load type 'MediatR.IRequest 1' from assembly 'MediatR, Version=10.0.0.0, Culture=neutral, PublicKeyToken=bb9a41a5e8aaa7e2' `

Any idea what's going on ?

vRITHNER:
Same issue for me with an AspNet.Core 6.0 app

jbogard:
I've got it running on .NET 6, is there a repro? Are you using the DI package?

jbogard:
And the exception message is correct - those types are in a separate MediatR.Contracts assembly as mentioned in the release notes. But user code shouldn't care about that in scenarios I looked at.

vRITHNER:
txs for coming back. Sadly, I can't send you the project cause it's a customer code.
I'll try to reproduce it in a total separate poc project and I'll come back to you..
Txs

jbogard:
If you check your stack trace, do you have code looking for the `IRequest` type from the MediatR assembly?

bend:
@jbogard  I don't see that in the stacktrade. However, I have the issue in multiple projects.
  [CS7069] Reference to type 'INotification' claims it is defined in 'MediatR', but it could not be found

I tried to force the update by install Mediatr10 in the project itself but I have the same issue.
Thanks

jbogard:
I don't know how to fix this without a minimal repro. I can only share that those request types are in a separate package/assembly - `MediatR.Contracts`, which is referenced by the `MediatR` package starting with version 10.

killergege:
Note that I have the same issue, but it happens when calling `MediatR.Extensions.FluentValidation.AspNetCore.AddFluentValidation()`.
Which is an external lib. 
The update probably shouldn't break that lib though...

Nobody mentions here what triggers the exception, so maybe we all have the same issue ?

In my case, the exception happen on the second line.
```
        services.AddMediatR(typeof(MyRequestHandler).GetTypeInfo().Assembly);
        services.AddFluentValidation(new[] { typeof(MyRequestHandlerValidator).GetTypeInfo().Assembly });
```
Removing it fixes the issue.

I have a very stripped down version of my project with the issue, but if the issue is in another lib, not sure this is interesting.

DMPatod:
[...]

PR #800, opened 2022-11-14
Adding type forwardings for types that were moved to the Contracts package

ydie22:
This pull request is a proposal to solve the issue described in https://github.com/jbogard/MediatR/issues/692

### DIFF
--- a/src/MediatR/TypeForwardings.cs
+++ b/src/MediatR/TypeForwardings.cs
@@ -1,0 +1,8 @@
+﻿using System.Runtime.CompilerServices;
+using MediatR;
+
+[assembly: TypeForwardedTo(typeof(IBaseRequest))]
+[assembly: TypeForwardedTo(typeof(IRequest<>))]
+[assembly: TypeForwardedTo(typeof(IRequest))]
+[assembly: TypeForwardedTo(typeof(INotification))]
+[assembly: TypeForwardedTo(typeof(Unit))]


jbogard:
Is anyone still having this issue? No more comments in many months. I don't want to pull this in if it's not a big issue anymore.

bend:
@jbogard On my side we still have this issue, this is a blocking us from upgrading to the new version of Mediatr (10 or 11)

jbogard:
It sounds this issue comes from transitive dependencies to MediatR:

- App A
  - MediatR 10.0
  - MediatR.Contracts 10.0
  - SomethingElse 1.0
     - MediatR 9.0

If that's the case, then this PR won't fix anything. MediatR contains breaking changes from 9.0 to 10.0 and 10.0 to 11.0. That `SomethingElse 1.0` package will break anyway.

The correct solution here is to fix SomethingElse package.

Unless there's some other issue I'm unaware of?

ydie22:
Let me explain what situation this PR is supposed to solve.
Say we start with the following situation:
- AppA
    - AppB.Contracts
        - MediatR 8.1
    - AppC.Contracts
        - MediatR 8.1
AppA works fine as everyone is using MediatR 8.1 where all types (contract-related and implementation) are still contained in a single assembly.
Now comes the MediatR version where contract types are separated in another package/assembly (which is btw, an excellent thing). The owners of AppC decide to migrate their MediatR version and they publish a new contracts package. Now AppA needs this new version of the contracts because there is a new feature exposed in the contracts. So they upgrade. The situation becomes:
- AppA
    - MediatR 11.0
    - AppB.Contracts
        - MediatR 8.1
    - AppC.Contracts
        - MediatR.Contracts 1.0
Notice that to be able to use the types in AppC.Contracts, AppA must now reference directly MediatR 11.0. And this is where things break. The package version will resolve to MediatR 11.0 and this one will be loaded at runtime. However, when AppB.Contracts was compiled, the MediatR interfaces used on request types it contains were present in the MediatR.dll (version 8.1). And now suddenly, they're no longer there as expected, because MediatR 11.0 does not contain them, they're in MediatR.Contracts. The runtime is lost, it cannot load the request types defined in AppB.Contracts.
However, if MediatR v11.0 would contain a hint to the runtime that the types he is looking for have moved, he would look at that additional place, and that is exactly what the PR does: it redirects the runtime to the new location where the types have moved. As AppB.Contracts and AppC.Contracts only contain contract definitions, they don't break; they are the kind of packages that MediatR.Contracts typically was designed for.

One could argue that AppB.Contracts should be upgraded to use MediatR.Contracts 1.0 also, but this is irrealistic once you start to have tens of services maintained by different teams. Such a forced synchronized upgrade is then really painful.

jbogard:
Ah that makes sense, to prevent forced upgrade of all services all at once.

ydie22:
@jbogard , thank you for merging this PR, it has saved us a lot of time while upgrading!

PR #802, opened 2022-11-20
Using MS.Extensions.DI directly, removing all unneeded code, pulling in the MediatR package directly

jbogard:
Brrrrrrrr


davidroth:
@jbogard Just wanted to ask if you are aware of the consequences for simple injector (and maybe other containers).
Simple Injector is a "non-conforming" DI Container, meaning that it doesn't replace the built-in DI Container, but instead lives side-by-side the built-in container. Where the built-in container is used to resolve framework and third-party components, Simple Injector is used by the application developer to resolve their application components. 

For general discussion regarding conforming containers see from former discussions such as https://github.com/dotnet/aspnetcore/issues/30115, https://github.com/dotnet/aspnetcore/issues/29194, https://github.com/dotnet/aspnetcore/issues/28957, https://github.com/dotnet/aspnetcore/issues/14585, https://github.com/dotnet/aspnetcore/issues/8886 and https://github.com/aspnet/DependencyInjection/issues/334.


So this change would break support for SimpleInjector. Realistically, there would be no way back (rather than forking).
Of course, its your project so are free to "dont care" and anyone is free to fork it😅

jbogard:
How would this break exactly? Doesn't the `Container` object implement `IServiceProvider`?

This doesn't force people to use the `ServiceCollection` stuff, it just bundles it together. The breaking change to the `Mediator` class is to change the dependency from my custom delegate type of `ServiceFactory` to the `System.ComponentModel` type `IServiceProvider`.

davidroth:
> This doesn't force people to use the ServiceCollection stuff, it just bundles it together.

Ah, I see.  Thanks for the clarification. I originally thought it requires using the frameworks DI Container (IServiceCollection).
But that is not the case as you said. By using the following registration we can still control the registration in the composition root:

```csharp
Container.RegisterInstance(typeof(IMediator), new Mediator(Container.GetInstance));
```

> Doesn't the Container object implement IServiceProvider?

It does currently, yes. May be [removed in the future though](https://github.com/simpleinjector/SimpleInjector/issues/874). But then we can still write a simple adapter:

```csharp

Container.RegisterInstance(typeof(IMediator), new Mediator(new ServiceProviderAdapter { Container = container }));

public class ServiceProviderAdapter : IServiceProvider
{
    public Container Container {get;set;}
    public object GetService(Type type) => this.Container.GetInstanceOrNull(type);
}
```
So everything seems to be good :-)

jbogard:
Yep, you should be able to write a simple adapter. MediatR used to use CommonServiceLocator years and years ago so this is a bit of return to form. I don't plan on altering anything with MediaR internals to force using its registration. In fact, I'm looking at adding factory objects to make it a bit more explicit about instantiating things like behaviors.

PR #821, opened 2023-01-14
Added target framework to netstandard2.0

maniglia:
Please consider adding netstandard2.0 as a target framework if you can.

### DIFF
--- a/src/MediatR/IStreamRequestHandler.cs
+++ b/src/MediatR/IStreamRequestHandler.cs
@@ -1,7 +1,5 @@
 using System.Collections.Generic;
 using System.Threading;
-using System.Threading.Tasks;
-using System.Runtime.CompilerServices;
 
 namespace MediatR;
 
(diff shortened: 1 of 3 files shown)


jbogard:
I...don't understand how this compiles. The docs say `IAsyncEnumerable` only supports `netstandard2.1` and above:

image

I'll do some more digging but if it's working with `netstandard2.0` then I'll just downgrade to that altogether.

jbogard:
Welp it gets pulled in with the Microsoft.Bcl.AsyncInterfaces package (which I did not know existed). Derp. We can just target `netstandard2.0` directly.

jbogard (changes requested):

jbogard (approved):

maniglia:
Unfortunately I have some problems compiling the project, i get the following message:
_MinVer : error MINVER1007: "git" is not present in PATH._
Even though I installed git and entered the environment variable.

jbogard (approved):

PR #831, opened 2023-02-05
Parallelization for notifications

RomanKreisel:
Hi all,

this is less of a PR but more of a request to integrate parallelization for notifications directly into MediatR.
I can only imagine, that there are some concerns why this hasn't been done yet, since even a (unnecessary?) complex sample was given how every consumer of MediatR could do this himself.
Nevertheless this seems to be a feature everyone using notifications is wondering why it's missing - including the famous Nick Chapsas (https://www.youtube.com/watch?v=2TT3suofNlo&t=7m11s) ;)

As I wasn't able to find any discussion about it though, please see this as a starting point for it. I'm absolutely open to further improve/extend this PR.


jbogard:
This has been discussed a few times, but the long and short of it is the implementation in the docs is STRICTLY to enable the sample app. It is not how I would design it if it were built-in.

PR #844, opened 2023-02-15
Use actual request type in new 'Send' method

remcoros:
See this old issue for background: https://github.com/jbogard/MediatR/issues/124

```csharp
var requests = new IRequest[] { new Req1(), new Req2(), new Req3() };
foreach (var req in requests)
{
  await _mediator.Send(req);
}
```

It will try to resolve handlers of type "IRequest" instead of using the concrete type of the given request.

sorry I didn't catch that yesterday 😞 let me know if I should add a test

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -63,7 +63,7 @@ public class Mediator : IMediator
             throw new ArgumentNullException(nameof(request));
         }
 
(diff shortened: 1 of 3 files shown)


jbogard:
Can you add a test for this? So I don't break it again.

jbogard:
I'll wait a bit to see if there are any other bugs before making a release for this.

PR #855, opened 2023-02-22
Partially fix #819

### LINKED ISSUE #819: (Bug) Contravariant handler is not resolved from Microsoft.DI service provider
## Related
https://stackoverflow.com/questions/54922190/microsoft-dependency-injection-and-contravariance

Microsoft wont fix: https://github.com/aspnet/DependencyInjection/issues/453

## Reproduction
https://github.com/tvardero/mediatr-contravariance

### Contracts
```csharp
record PaintCarCommandBase : INotification
{
    public required string Color { get; init; }
}

record PaintTeslaCommand : PaintCarCommandBase { }

record PaintToyotaCommand : PaintCarCommandBase
{
    public required string SecondaryColor { get; init; }
}
```

### Handlers
```csharp
abstract class CommandHandlerBase<TCommand> : INotificationHandler<TCommand>
where TCommand : INotification
{
    public abstract Task Handle(TCommand command, CancellationToken cancel);
    // do some logging here, command validation, etc
    // or maybe make this CommandHandlerBase reused also by RabbitMq
}

class PaintCarCommandHandler : CommandHandlerBase<PaintCarCommandBase>
{
    public override Task Handle(PaintCarCommandBase command, CancellationToken cancel)
    {
        Console.WriteLine($"Painting car using primary color: {command.Color}");

        if (command is PaintToyotaCommand toyotaCommand)
        {
            Console.WriteLine($"Painting spoilers using secondary color: {toyotaCommand.SecondaryColor}");
        }

        return Task.CompletedTask;
    }
}
```

### The Code
```csharp
serviceCollection.AddMediatR(Assembly.GetExecutingAssembly());

using var serviceProvider = serviceCollection.BuildServiceProvider();

var toyotaPainting = new PaintToyotaCommand { Color = "black", SecondaryColor = "red" };
var teslaPainting = new PaintTeslaCommand { Color = "bright pink" };

var mediator = serviceProvider.GetRequiredService<IMediator>();

mediator.Publish(toyotaPainting); // <-- unfortunately does nothing
mediator.Publish(teslaPainting); // <-- unfortunately does nothing

var h1 = serviceProvider.GetService<INotificationHandler<PaintCarCommandBase>>(); // <-- not null, OK
var h2 = serviceProvider.GetService<INotificationHandler<PaintToyotaCommand>>(); // <-- Null
var h3 = serviceProvider.GetService<INotificationHandler<PaintTeslaCommand>>(); // <-- Null

Console.ReadLine();
```

## How MediatR could fix that
I have investigated some source code and found something. Lets say we are handling Toyota painting.

This line creates NotificationHandlerWrapper with type of PaintToyotaCommand.
https://github.com/jbogard/MediatR/blob/09dbb3e8c5c58b6e949e06e57755211571efda48/src/MediatR/Mediator.cs#L110

Later, wrapper attempts to find handlers of type `INotificationHandler<PaintToyotaCommand>`:
https://github.com/jbogard/MediatR/blob/09dbb3e8c5c58b6e949e06e57755211571efda48/src/MediatR/Wrappers/NotificationHandlerWrapper.cs#L26

And as we saw earlier, this returns null as Microsoft.DI does not support contravariance.

The fix could be simple, find INotificationHandler for base class too (recursive), if it is also marked with `INotification` interface.

tvardero:


tvardero:
@jbogard, I have added one unit test that covers the situation

src/MediatR/Internal/TypeExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// <param name="filter">The delegate that compares type against <paramref name="filterCriteria"/>.</param>
+    /// <param name="filterCriteria">The search criteria that determines whether an type should be included in the returned collection.</param>
+    /// <exception cref="ArgumentNullException">Either <paramref name="type"/> of <paramref name="filter"/> is null.</exception>
+    /// <returns>Collection of types the provided <paramref name="type"/> inherits explicitly or implicitly.</returns>
+    public static IEnumerable<Type> GetBaseTypes(this Type type, TypeFilter filter, object? filterCriteria)
+    {
+        if (type == null) throw new ArgumentNullException(nameof(type));
+        if (filter == null) throw new ArgumentNullException(nameof(filter));
+
+        var collection = new Stack<Type>();
```
asimmon, line 20: I think you could yield the results instead

src/MediatR/Wrappers/NotificationHandlerWrapper.cs
```diff
@@ -21,9 +22,18 @@ public class NotificationHandlerWrapperImpl<TNotification> : NotificationHandler
         Func<IEnumerable<NotificationHandlerExecutor>, INotification, CancellationToken, Task> publish,
         CancellationToken cancellationToken)
     {
-        var handlers = serviceFactory
-            .GetServices<INotificationHandler<TNotification>>()
-            .Select(static x => new NotificationHandlerExecutor(x, (theNotification, theToken) => x.Handle((TNotification)theNotification, theToken)));
+        var handlers = typeof(TNotification)
+            // For each of base types of TNotification, including TNotification itself
+            .GetBaseTypes(static (t, _) => t.GetInterfaces().Any(static i => i == typeof(INotification)), null)
```
asimmon, line 27: I suspect this to have an impact on performance, the only way to be sure would be to run some benchmarks to measure CPU time but also memory usage.

src/MediatR/Wrappers/NotificationHandlerWrapper.cs
```diff
@@ -21,9 +22,18 @@ public class NotificationHandlerWrapperImpl<TNotification> : NotificationHandler
...
+            .Append(typeof(TNotification))
+
+            // Resolve all INotificationHandler<...> for the type from service provider
+            .Select(static t => typeof(INotificationHandler<>).MakeGenericType(t))
+            .SelectMany(serviceFactory.GetServices)
```
asimmon, line 32: Not sure, but is there any chance that we end up with duplicate handlers?

src/MediatR/Wrappers/NotificationHandlerWrapper.cs
tvardero: I was unable to run benchmark that is already present in project.

P.S. I'm not sure that other containers cache their service evaluations, MS.Ext.DI does it per each type that was requested.
Also we do have our internal caching of handler executors, per each Mediator instance. Would Mediator be registered as scoped/singleton, the cache will help to reduce the impact.

tvardero:
Sometimes I think that the issue should not be resolved on Mediator end, as some other containers (not MS.Ext.DI, but for example Lamar) does support contravariance and it was working before the proposed patch.

This patch should only be applied to MS.Ext.DI and not all other containers, @asimmon could you suggest me how to do such?

cc @jbogard

jbogard:
Yeah I'm more inclined to let 3rd-party containers take care of this. They'll do the job much better than I can, and I've already taken on a bit much by adding the scanning for handlers.

tvardero:
For MS.Ext.DI I have already (re-)opened the issue: https://github.com/dotnet/runtime/issues/82372
They already said "wontfix" back in 2016, but I have hope for such change today.

asimmon:
@tvardero Before 12.x, MediatR was container-agnostic_-ish_. If you look at [v11.1.0/src/MediatR/Mediator.cs](https://github.com/jbogard/MediatR/blob/v11.1.0/src/MediatR/Mediator.cs#L25), there was a `ServiceFactory` acting as a DI abstraction layer.

In 12.x, `IServiceProvider` from Ms.Ext.DI is the built-in way of resolving dependencies. I **personally** think this a valid approach as ASP.NET Core, hosting extensions and many Microsoft and third-party libraries rely on this by default. It's part of the .NET ecosystem. Too ensure projects longevity and smooth migrations to more recent .NET versions in the next 5 years at least, developers should rely on Ms.Ext.DI.

tvardero:
> I would expect developers to use IRequest and IRequestHandler instead.

Isn't behavior of the `IRequest` (or we could say `IRequest<void>` / `IRequest<Unit>`) and it's `IRequestHandler` will be the same as with `INotification` and it's handler? Except for exceptions handling probably, INotification should work as fire-and-forget. Anyway, the #819 was about resolving a handler. I haven't done any changes to IRequestHandler resolution with MS.Ext.DI, but I assume there will be the same issue with contravariance.

jbogard:
> Isn't behavior of the `IRequest` (or we could say `IRequest<void>` / `IRequest<Unit>`) and it's `IRequestHandler` will be the same as with `INotification` and it's handler? Except for exceptions handling probably, INotification should work as fire-and-forget.

No, it's not. There is an expectation of a single handler for a single request, and 0-n handlers for a single notification. This is enforced by how we use the container, it will throw an exception without a request handler, and do nothing without a notification handler.

Generic variance makes sense in the context of notification handlers and behaviors. NOT in request handlers. Those should be closed generics, never open.

src/MediatR/Internal/ObjectTypeEqualityComparer.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+
+namespace MediatR.Internal;
+
+public class ObjectTypeEqualityComparer : IEqualityComparer<object?>
```
asimmon, line 5: The class `Type` already implement `Equals` and `GetHashCode` so I don't think this is necessary.
Also if you look at its `Type.Equals` implementation it does `object.ReferenceEquals(this.UnderlyingSystemType, o.UnderlyingSystemType)` which seems better.

asimmon:
So, I did the benchmarks by isolating the path that resolve the `INotificationHandler<T>` (so just before we create the wrappers) to measure the impact of your changes.

**TL;DR: it is 10x slower and consumes also 10x more memory.**

```
|      Method |       Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|------------ |-----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| BeforePR855 |   156.5 ns |  1.79 ns |  1.68 ns |  1.00 |    0.00 | 0.0162 |     136 B |        1.00 |
|  AfterPR855 | 1,575.5 ns | 14.58 ns | 13.64 ns | 10.07 |    0.13 | 0.1659 |    1392 B |       10.24 | <--- This PR
```

The benchmark code [can be found here](https://gist.github.com/asimmon/b7c74508be9999d2f93ca294ff47a3de).

While I completely agree that [you probably don't need to worry about MediatR's performance](https://jimmybogard.com/you-probably-dont-need-to-worry-about-mediatr/), in my opinion this PR **is a big no no**. I'd like your opinion about this @jbogard  if you're keen.

If I had to summarize:

* 10x slower and 10x more allocations due to reflection
* Usage of `SelectMany()` on `IServiceProvider.GetServices` which frankly seems odd
* Handler order disrupted. Usually `IServiceProvider.GetServices` takes care of returning services in their registration order, but here we interfere with that.

src/MediatR/Internal/ObjectTypeEqualityComparer.cs
tvardero: This scenario then will result in duplication:
```csharp
class Foo {}
class Bar : Foo {}

services.AddTransient<INotificationHandler<Foo>, FooHandler>();
services.AddTransient<INotificationHandler<Bar>, FooHandler>();

mediator.Publish(new Bar());
```

PR #857, opened 2023-02-23
Skip unnecessary BCL package reference if targeting .NET 6 or later

asimmon:
As mentioned in [#852](https://github.com/jbogard/MediatR/pull/852#issuecomment-1439443067), I would like to introduce multi-targeting **only to skip an unnecessary package reference** (`Microsoft.Bcl.AsyncInterfaces`) on .NET 6 and later, because it's already included by default.

A (positive?) side effect of targeting `net6.0` is that a few warnings appeared in existing code because of new nullable reference types annotations. For instance, `IServiceProvider.GetService(Type serviceType)`:

* Returns `object` in `netstandard2.0`
* Returns `object?` in `net6.0`

At runtime, the behavior is the same. It's just that Microsoft added more hints to prevent null reference exceptions. I had to fix these warnings, and that explains the number of changes.

While fixing these warnings in `Mediator.cs`, I also added:
* Missing `InvalidOperationException` if we can't create the wrapper for stream requests handlers (instead of null-ref ex),
* More consistency in how wrappers are cached in their respective concurrent dictionaries (static delegates everywhere for better memory usage),
* More consistency in `InvalidOperationException` messages

### DIFF
--- a/src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
+++ b/src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
@@ -1,6 +1,7 @@
 namespace MediatR.Pipeline;
 
 using Internal;
+using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -71,7 +72,7 @@ public class RequestExceptionActionProcessorBehavior<TRequest, TResponse> : IPip
         var exceptionActionInterfaceType = typeof(IRequestExceptionAction<,>).MakeGenericType(typeof(TRequest), exceptionType);
         var enumerableExceptionActionInterfaceType = typeof(IEnumerable<>).MakeGenericType(exceptionActionInterfaceType);
 
-        var actionsForException = (IEnumerable<object>)_serviceProvider.GetService(enumerableExceptionActionInterfaceType);
+        var actionsForException = (IEnumerable<object>)_serviceProvider.GetRequiredService(enumerableExceptionActionInterfaceType);
 
         return HandlersOrderer.Prioritize(actionsForException.ToList(), request)
             .Select(action => (exceptionType, action));
--- a/src/MediatR/MediatR.csproj
+++ b/src/MediatR/MediatR.csproj
@@ -4,7 +4,7 @@
     <Authors>Jimmy Bogard</Authors>
     <Description>Simple, unambitious mediator implementation in .NET</Description>
     <Copyright>Copyright Jimmy Bogard</Copyright>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
     <Nullable>enable</Nullable>
     <Features>strict</Features>
     <PackageTags>mediator;request;response;queries;commands;notifications</PackageTags>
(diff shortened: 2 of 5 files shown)


asimmon:
😕Seems like there is a flaky test that relies on timing (`MediatR.Tests.NotificationPublisherTests.Should_handle_sequentially_by_default`). I guess that this is to test parallel notification handlers against sequential notification handlers, but on GitHub agents this can be unreliable.

src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
```diff
@@ -71,7 +72,7 @@ private IEnumerable<(Type ExceptionType, object Action)> GetActionsForException(
         var exceptionActionInterfaceType = typeof(IRequestExceptionAction<,>).MakeGenericType(typeof(TRequest), exceptionType);
         var enumerableExceptionActionInterfaceType = typeof(IEnumerable<>).MakeGenericType(exceptionActionInterfaceType);
 
-        var actionsForException = (IEnumerable<object>)_serviceProvider.GetService(enumerableExceptionActionInterfaceType);
+        var actionsForException = (IEnumerable<object>)_serviceProvider.GetRequiredService(enumerableExceptionActionInterfaceType);
```
jbogard, line 75: What are the implications of changing this to GetRequiredService?

jbogard (commented):
Were the additional null checks a side effect of .NET 6?

src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
asimmon: [GetRequiredService](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getrequiredservice?view=dotnet-plat-ext-6.0#microsoft-extensions-dependencyinjection-serviceproviderserviceextensions-getrequiredservice(system-iserviceprovider-system-type)) throws a `InvalidOperationException` when the requested service type was **not** registered.

[GetService](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getservice?view=dotnet-plat-ext-6.0#returns) returns `null` when the requested service was not registered. This behavior is not framework-specific. In the current state of the master branch, this could _eventually_ happen. The following cast to (`IEnumerable<object>`) would have thrown a null-ref exception.

It's either we use `GetRequiredService` or we suppress the nullable warning with `!` to say that we are 100% sure that there will be a registration for the requested type.

asimmon:
> Were the additional null checks a side effect of .NET 6?

Yes. Basically it adds more precise hints on the existing runtime behavior. I think it's a positive thing, it puts in evidence paths that could eventually throw a null-reference exception. `GetService` is a good example. In the current master state, it could return null but there's no warning whatsoever.

src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
jbogard: I'm going to add a unit test for this scenario. `GetService` should *never* return null for enumerables, it would be an empty list. But I don't know what the behavior would be for `GetRequiredService` for enumerables, would it throw if there aren't any registrations? I wouldn't think so, but work verifying.

jbogard (approved):

src/MediatR/Pipeline/RequestExceptionActionProcessorBehavior.cs
asimmon: Oh it's a little bit late but there is also a built-in [IServiceProvider.GetServices(serviceType)](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceproviderserviceextensions.getservices?view=dotnet-plat-ext-6.0) which returns an `IEnumerable<T>`, so basically it returns an empty list if there's no registration for `T`.

PR #860, opened 2023-02-23
AddOpenBehavior(type) now supports stream behaviors without duplicate registrations

asimmon:
What's changed:

* `MediatRServiceConfiguration.AddOpenBehavior(type)` assumed that the type parameter implemented `IPipelineBehavior<,>`. It now supports `IStreamPipelineBehavior<,>` and throws an `InvalidOperationException` the type parameter is not _actually_ an open behavior. Added multiple tests.

* `TryAddTransientExact` was replaced by the built-in [`IServiceCollection.TryAddEnumerable`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.extensions.servicecollectiondescriptorextensions.tryaddenumerable?view=dotnet-plat-ext-6.0) that does the exact same thing, i.e. skip a registration if there's already a service descriptor that exactly match the service type AND implementation type. Modified an existing test to prove that everything behaves as before.

* `MediatRServiceConfiguration.BehaviorsToRegister` consumption now use `TryAddEnumerable` too. Covered by the modified test from the previous point.

* Apply the [11.x to 12.x migration guide](https://github.com/jbogard/MediatR/wiki/Migration-Guide-11.x-to-12.0#rolling-back-generic-constraints) in test classes: several test open behaviors were not using the right generic constraint, which is now `TRequest : notnull`.

### DIFF
--- a/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using MediatR;
 using MediatR.NotificationPublishers;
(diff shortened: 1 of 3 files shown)


jbogard (approved):

asimmon:
Hi @jbogard would you consider releasing these changes in a new patch?

PR #894, opened 2023-04-24
This PR  makes the marker interface `INotification` optional.

vvdb-architecture:
`INotification` is an unnecessary constraint. It's absence does not prevent the correct functioning of `INotificationHandler`/`INotificationPublisher`.

This avoids the creation of a specific type for each notification type, and allows us to define handlers using value tuples:

~~~csharp
public class Entity
{
}

public enum Operation
{
  Create,
  Update
}

public class EntityEventHandler : INotificationHandler<(Entity Entity, Operation Operation)>
{

    public Task Handle((Entity Entity, Operation Operation) notification, CancellationToken cancellationToken)
    {
    }
}

~~~

... and then call them using a syntax like:

~~~csharp
        mediator.Publish((entity, Operation.Create));
~~~

### DIFF
--- a/samples/MediatR.Examples.PublishStrategies/CustomMediator.cs
+++ b/samples/MediatR.Examples.PublishStrategies/CustomMediator.cs
@@ -7,11 +7,11 @@ namespace MediatR.Examples.PublishStrategies;
 
 public class CustomMediator : Mediator
 {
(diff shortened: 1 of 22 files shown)


jbogard:
The `INotification` interface is also there to prevent mistakes. Not publishing requests or sending notifications. I don't want to go back on that just to enable tuples.

PR #977, opened 2023-11-26
IPublisher added to contracts

VarunSaiTeja:
I have a project for data persistence.
When I saved data with UOW, I want to publish an event, Since its persistence project I dont want to use MediatR full package in my persistence layer.

So I thought it would be nice to have IPublisher in contracts.


jbogard:
That's not what the Contracts package is for. It's for sharing request/response and notification contracts across systems.

PR #988, opened 2023-12-26
Adding support for Order as an integer property.

### LINKED ISSUE #987: Allow for an optional "Order" parameter to be set on implementations of pre, post, and exception handlers.
(this issue exists so I have a branch name to create my PR against)

There are many examples of users of MediatR wanting to execute pre & post-request handlers in order.  For instance, running a validator to fail early before doing some heavier lookups or pre-populating with downstream processors.  Many DI containers support adding these handlers in a specific order, but this means you cannot use a type scanner to auto-populate the list.

Newer versions of C# allow for auto-implementation of interface members.  As far as I can tell, this is an ideal way to implement the idea of a numeric "Order".  By adding an integer "Order" field, defaulting it to 0 in the interface, and ordering by "Order" when executing Pre/Post/Exception handlers, a developer could optionally support this feature in a native way with very little overhead.

bryanboettcher:
Should resolve #987.

Signed off by: Bryan Boettcher <bryan.boettcher@gmail.com>

### DIFF
--- a/src/MediatR/Internal/HandlerExtensions.cs
+++ b/src/MediatR/Internal/HandlerExtensions.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+
+namespace MediatR.Internal;
+
+/// <summary>
+/// Contains utility methods for working with implementations of request
+/// handlers.
+/// </summary>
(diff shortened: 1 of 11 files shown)


bryanboettcher:
I am still interested in trying to solve this.

PR #989, opened 2023-12-29
Fix AutoRegisterRequestProcessors to include all implementations

hisuwh:
This is how this previously worked.  When this option was restored the `addIfAlreadyExists` property was mistakenly set to false.

I have updated the tests and changed this property.

### DIFF
--- a/src/MediatR/Registration/ServiceRegistrar.cs
+++ b/src/MediatR/Registration/ServiceRegistrar.cs
@@ -23,8 +23,8 @@ public static class ServiceRegistrar
 
         if (configuration.AutoRegisterRequestProcessors)
         {
-            ConnectImplementationsToTypesClosing(typeof(IRequestPreProcessor<>), services, assembliesToScan, false, configuration);
-            ConnectImplementationsToTypesClosing(typeof(IRequestPostProcessor<,>), services, assembliesToScan, false, configuration);
+            ConnectImplementationsToTypesClosing(typeof(IRequestPreProcessor<>), services, assembliesToScan, true, configuration);
+            ConnectImplementationsToTypesClosing(typeof(IRequestPostProcessor<,>), services, assembliesToScan, true, configuration);
         }
 
         var multiOpenInterfaces = new List<Type>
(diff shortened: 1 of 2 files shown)


asimmon:
Hi @jbogard, I agree with @hisuwh's claim that there is a bug in the way the new `AutoRegisterRequestProcessors` boolean flag works. Here's the simplest repro case I've got. The pre/post processors are simply not executed. I believe this is the most trivial use of processors.

```csharp
// <PackageReference Include="MediatR" Version="12.2.0" />
// <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />

using MediatR;
using MediatR.Pipeline;
using Microsoft.Extensions.DependencyInjection;

var services = new ServiceCollection().AddMediatR(x =>
{
    x.AutoRegisterRequestProcessors = true;
    x.RegisterServicesFromAssemblyContaining<Program>();
});

await using var serviceProvider = services.BuildServiceProvider();
await serviceProvider.GetRequiredService<IMediator>().Send(new MyRequest());

public class MyRequest : IRequest;

public class MyRequestHandler : IRequestHandler<MyRequest>
{
    public Task Handle(MyRequest request, CancellationToken cancellationToken) => Task.CompletedTask;
}

public class MyRequestProcessor : IRequestPreProcessor<MyRequest>, IRequestPostProcessor<MyRequest, Unit>
{
    public Task Process(MyRequest request, CancellationToken cancellationToken) => Task.CompletedTask;
    public Task Process(MyRequest request, Unit response, CancellationToken cancellationToken) => Task.CompletedTask;
}
```

It's very unfortunate as we were about to leverage this pre/post processor in one of my team's features. Our only option right now is to downgrade MediatR to 12.0.1, which involves not benefiting from the recent improvements and other bug fixes.

bryanboettcher:
+1 for "this is impacting me"

hisuwh:
@jbogard do you want me to change anything on this or add any more tests?  I'm happy to do so.
Otherwise, can we get this in?

jbogard:
That scanning stuff is code I copied from some other DI projects, so you can see why I don't really want to copy more behaviors from DI containers because that code is...complicated

hisuwh:
@jbogard when are you planning new release with this in please 🙏🥺

asimmon:
I thought this change would fix this very simple usecase but it still doesn't work with 12.3.0:

https://github.com/jbogard/MediatR/pull/989#issuecomment-1883574379

PR #1015, opened 2024-03-26
Moved interfaces to Contracts

harrypooni:
Extracted Interfaces to Contracts to provide abstraction to multi project solutions.

### DIFF
--- a/src/MediatR/MediatR.csproj
+++ b/src/MediatR/MediatR.csproj
@@ -38,4 +38,8 @@
     <PackageReference Include="MinVer" Version="4.3.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\MediatR.Contracts\MediatR.Contracts.csproj" />
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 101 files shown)


Mayron:
@harrypooni as discussed in the issue, I have created my own pull request: https://github.com/jbogard/MediatR/pull/1020

I think your PR as the issue of including `MediatR.Contracts` as a project reference when it should only depend on it via a NuGet package.

The PR I created tries to create the other abstractions/interfaces as additional NuGet packages and MediatR could eventually be updated to include them as dependencies instead of directly referencing the projects.

src/MediatR/MediatR.csproj
```diff
@@ -38,4 +38,8 @@
     <PackageReference Include="MinVer" Version="4.3.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\MediatR.Contracts\MediatR.Contracts.csproj" />
+  </ItemGroup>
+
```
Mayron, line 44: This is the issue I'm referring to.

jbogard:
No, I really really really don't want to have to maintain all these packages and projects when I've never needed to have them broken out as such. I do not have any scenarios where breaking them out further would enable some scenario or solve some dependency problem.

Issue #1014, opened 2024-03-26
Seperate Interfaces to Contract project to provide abstraction layer


harrypooni:
Move interfaces away from implementation provides flexibility when projects are separated in consumer code base.

Mayron:
To add to this, I was thinking today whether it would be possible to have `IRequestHandler` in another package (or in the `MediatR.Contracts` package if that's possible (probably not, I'd guess). 

The reason is because I have a `Presentation` layer of my architecture that uses `IMediator` to send requests, and the handlers are in a different project, as part of the `Application` layer, and implement `IRequestHandler`. But I want to avoid being able to inject `IMediator` into those handlers to prevent handlers sending out more requests.

I wish the handlers didn't have a reference to `IMediator` because I think it would solve this issue. They only need `IRequestHandler` but that's not in the Contracts package unlike `IRequest`. I found this a bit unusual.

khellang:
Physically separating types to enforce architectural styles sound like a slippery slope. What you're requesting sounds like a few types per assembly, which seems excessive. Why don't you write some architecture tests or a Roslyn analyzer to enforce your requirements instead?

> Is this something that could be done in a way that's backwards compatible with previous versions of MediatR?

No, the old assembly/package would need to still reference the new assembly/package and apply a [`TypeForwardedToAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.typeforwardedtoattribute?view=net-8.0) in order to make it backwards compatible, but that seems to defeat your requirement(s).

harrypooni:
Any objections to the pr being approved?

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Mayron:
@harrypooni I would very much appreciate just one other project containing the abstractions for IMediator and ISender instead of including them in the Contracts project because this would solve my issue mentioned above. The issue with sending nested requests inside handlers has been mentioned a number of times on previous issues and blogs/posts across the internet, and having this other project is a very simple change, especially since you've already done most of the work in this PR.

Mayron:
[...]

PR #1020, opened 2024-04-10
#1014 Pipeline and Services Abstractions

Mayron:
closes: #1014 

I have created 2 new abstraction packages/projects: `MediatR.Pipeline` and `MediatR.Services` (feel free to rename them). I have left `MediatR` and `MediatR.Contracts` untouched and have included `MediatR.Contracts` as a NuGet package dependency on the new packages.

Eventually, `MediatR` can include these 2 new NuGet packages as dependencies and remove the duplicate interfaces from its source code.

I left out the interfaces used in `MediatR` inside the `MediatR.Pipeline` namespace because they're more coupled with pipeline-related concepts and I don't think these are being requested from the initial issues raised. If more work is required, let me know or feel free to update this PR.

---

The reason for these additions is to allow for more architectural flexibility when using MediatR (which also solves my initial issue). For more information, please refer to the [linked issue](https://github.com/jbogard/MediatR/issues/1014).

**TL;DR:** I can have one project depend on the `MediatR.Pipeline` NuGet package to store implementations of handlers and behaviors without being allowed to use services to send nested requests, and I can have another project depending on the `MediatR.Services` NuGet package for sending requests without implementing handlers.

### DIFF
--- a/MediatR.Pipeline/INotificationHandler.cs
+++ b/MediatR.Pipeline/INotificationHandler.cs
@@ -1,0 +1,40 @@
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace MediatR;
+
+/// <summary>
+/// Defines a handler for a notification
+/// </summary>
+/// <typeparam name="TNotification">The type of notification being handled</typeparam>
+public interface INotificationHandler<in TNotification>
+    where TNotification : INotification
+{
+    /// <summary>
+    /// Handles a notification
+    /// </summary>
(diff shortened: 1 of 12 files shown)


jbogard:
See discussion, I'm not interested in splitting packages out any further unless there are demonstrated needs for multiple applications or published libraries (i.e., beyond a single application) to reference a subset of MediatR features AND having the said features in a single library is preventing some capability.

PR #1032, opened 2024-05-17
Set TargetFrameworks solely to "netstandard2.0"

marco-carvalho:
Changed the `<TargetFrameworks>` from `netstandard2.0;net6.0` to just `netstandard2.0`. It shouldn't affect users since `netstandard2.0` works with .NET 6.


RVPE:
https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting

Issue #1047, opened 2024-06-27
Generic Handler implementation only handles single generic argument and breaks existing implementations 


hisuwh:
Support for Generic Handlers has been added in v12.3.0 in this pr: #1013

However, this is not opt-in and only supports a single generic argument.

So a setup like this is incompatible:

```csharp
public class MyRequest<T1, T2> : IRequest<T2>

public class MyRequestHandler<T1, T2> : IRequestHandler<MyRequest<T1, T2>, T2>
```

Also rather than simply not matching these requests this throws an exception on startup meaning this can't even fallback to existing code I have to handle these more complex generic handlers.

I suggest:
- if it cannot find/close a handler it should ignore it (as per standard behaviour) - and not throw an exception!
- there should be an option to opt-out
- and/or this should be extended to handle multiple generic arguments

hisuwh:
@jbogard any chance of a 12.2.1 to fix https://github.com/jbogard/MediatR/pull/989

If and when I get some time I can look at the above as I have implemented generic handlers in a few projects.

zachpainter77:
@jbogard  I have a fix for the above issue.. My only question is what limitations should be present?  What I mean is should we let the user define handlers that have 10 type parameters, each with no constraint? My solution is to allow more than one generic type parameter but throw errors if the max number of generic type params is exceeded OR the max number of concrete class combinations is exceed OR if a timeout occurs.  My question is.. what should those limitations be?

I'm also trying to figure out a way to test these scenarios since even defining a class that should throw an exception does so for all tests that call `AddMediatR`.  

So, any discussion around this is welcome.  I think I'm just going to create a PR and let discussion take place there.

hisuwh:
Is there a way to define a sensible limit internally, but then allow for this to be configurable?

zachpainter77:
yes, I think there is a way to do this.. I will set up some configuration options with default values that can be overridden when calling `AddMediatR`.. I will have a PR coming today. We can probably discuss what is sensible in terms of default limitations in that thread.

jbogard:
This fix is on the MyGet feed: https://myget.org/gallery/mediatr-ci

Can you verify that this fixes the issue?

zachpainter77:
![image](https://github.com/user-attachments/assets/7d0e1e5e-d5e9-4e52-91e7-2c4362e8654e)
 Seems to work with my test project. However I had to use this myget feed url in order to install the package via nuget package manager in visual studio.. 
 
 
[...]

Issue #1038, opened 2024-06-07
v12.3.0 Breaking Change


bsal649:
After updating from 12.2.0 to 12.3.0, I'm getting a runtime exception during app instantiation.

The code:
```csharp
// Register MediatR
Assembly[] assemblies = Assembly.GetEntryAssembly()!.GetReferencedAssemblies()
                        .Select(Assembly.Load).Append(Assembly.GetCallingAssembly()).ToArray();

_ = services.AddMediatR(config => config.RegisterServicesFromAssemblies(assemblies)); // exception happens here
```

The exception:
```
Exception has occurred: CLR/System.ArgumentException
An exception of type 'System.ArgumentException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'The number of generic arguments provided doesn't equal the arity of the generic type definition.'
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at MediatR.Registration.ServiceRegistrar.<>c__DisplayClass5_0.<GetConcreteRequestTypes>b__2(Type type)
   at System.Linq.Enumerable.SelectListIterator`2.Fill(ReadOnlySpan`1 source, Span`1 destination, Func`2 func)
   at System.Linq.Enumerable.SelectListIterator`2.ToList()
   at MediatR.Registration.ServiceRegistrar.GetConcreteRequestTypes(Type openRequestHandlerInterface, Type openRequestHandlerImplementation, IEnumerable`1 assembliesToScan)
   at MediatR.Registration.ServiceRegistrar.AddAllConcretionsThatClose(Type openRequestInterface, List`1 concretions, IServiceCollection services, IEnumerable`1 assembliesToScan)
   at MediatR.Registration.ServiceRegistrar.ConnectImplementationsToTypesClosing(Type openRequestInterface, IServiceCollection services, IEnumerable`1 assembliesToScan, Boolean addIfAlreadyExists, MediatRServiceConfiguration configuration)
   at MediatR.Registration.ServiceRegistrar.AddMediatRClasses(IServiceCollection services, MediatRServiceConfiguration configuration)
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.AddMediatR(IServiceCollection services, MediatRServiceConfiguration configuration)
   at Microsoft.Extensions.DependencyInjection.ServiceCollectionExtensions.AddMediatR(IServiceCollection services, Action`1 configuration)
[...]

PR #1048, opened 2024-07-05
Add Support For Generic Handlers With Multiple Generic Type Parameters

zachpainter77:
This PR adds support for handlers that contain multiple generic type parameters.  

This PR is a direct fix for issues #1047 and #1038 

Example:

```csharp

public interface IZong{}
public class Zong : IZong {}
public interface IDong{}
public class Dong : IDong {}
public interface IPong { string Message? { get; } }
public class Pong : IPong
{
    string Message? { get; set; }
}

//generic request definition
public class GenericPing<TPong, TZong, TDong> : IRequest<TPong>
    where TPong : IPong
    where TZong : IZong
    where TDong : IDong
{
    public T? ThePong { get; set; }
}

//generic request handler
public class GenericPingHandler<TPong, TZong, TDong> : IRequestHandler<GenericPing<TPong, TZong, TDong>, TPong>
    where TPong : IPong
    where TZong : IZong
    where TDong : IDong
{
    public Task<TPong> Handle(GenericPing<TPong, TZong, TDong> request, CancellationToken cancellationToken) => Task.FromResult(request.ThePong!);
}

//usage
var pong = _mediator.Send(new GenericPing<Pong, Zong, Dong>{ Pong = new() { Message = "Ping Pong" } });
Console.WriteLine(pong.Message); //would output "Ping Pong"

```
The issue that this PR solves was caused from only attempting to close the first type parameter in a generic request handler implementation.

This PR fixes that by finding the types that close for all type parameter and storing those in a list of lists.  Then all possible combinations of concrete implementations are generated from the lists and registered with the service provider.

Something that might need some discussion and thought will be what kinds of limitations we should put around this feature so that users are not misusing the library and creating a deadlock during service registration.

Along with this PR I added these configuration options with default values to the Configuraton object..

```csharp
 /// <summary>
    /// Configure the maximum number of type parameters that a generic request handler can have. To Disable this constraint, set the value to 0.
    /// </summary>
    public int MaxGenericTypeParameters { get; set; } = 10;

    /// <summary>
    /// Configure the maximum number of types that can close a generic request type parameter constraint.  To Disable this constraint, set the value to 0.
    /// </summary>
    public int MaxTypesClosing { get; set; } = 100;

    /// <summary>
    /// Configure the Maximum Amount of Generic RequestHandler Types MediatR will try to register.  To Disable this constraint, set the value to 0.
    /// </summary>
    public int MaxGenericTypeRegistrations { get; set; } = 125000;

    /// <summary>
    /// Configure the Timeout in Milliseconds that the GenericHandler Registration Process will exit with error.  To Disable this constraint, set the value to 0.
    /// </summary>
    public int RegistrationTimeout { get; set; } = 15000;
```
For now these are the default values, but I would like some feedback regarding these values.

- If the user tries to register types that have more than 10 generic type arguments, the library will throw exception.
- if the user tries to register a type that has no generic type constraints the library will throw exception.
- if the user trues to register a type that has too many types that can close a specific parameter, the library will throw exception.
- if the user tries to register a handler and somehow gets past the previous checks then the library will check the total combinations and throw exception if that number is greater than the max.
- and last if the registration takes more time to complete than the timeout value in milliseconds the library will throw exception.

I've also added some pretty extensive test methods that confirm this is working as expected.

@jbogard , @hisuwh Please review this at your convenience. 

Thanks,

 - Zach
 
 Edit: I apologize for not getting this done sooner.. I have just started a new job and haven't had the time. :)
 
 Edit: Added an opt out functionality. 
 ```csharp
   services.AddMediatR(cfg =>
   {
       //opt out flag set
       cfg.RegisterGenericHandlers = false;
       cfg.RegisterServicesFromAssembly(assembly);
   });
```

### DIFF
--- a/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
@@ -70,6 +70,26 @@ public class MediatRServiceConfiguration
     public bool AutoRegisterRequestProcessors { get; set; }
 
     /// <summary>
+    /// Configure the maximum number of type parameters that a generic request handler can have. To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int MaxGenericTypeParameters { get; set; } = 10;
+
+    /// <summary>
+    /// Configure the maximum number of types that can close a generic request type parameter constraint.  To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int MaxTypesClosing { get; set; } = 100;
+
+    /// <summary>
+    /// Configure the Maximum Amount of Generic RequestHandler Types MediatR will try to register.  To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int MaxGenericTypeRegistrations { get; set; } = 125000;
+
+    /// <summary>
+    /// Configure the Timeout in Milliseconds that the GenericHandler Registration Process will exit with error.  To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int RegistrationTimeout { get; set; } = 15000;
+
+    /// <summary>
     /// Register various handlers from assembly containing given type
     /// </summary>
     /// <typeparam name="T">Type from assembly to scan</typeparam>
--- a/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
@@ -90,6 +90,11 @@ public class MediatRServiceConfiguration
     public int RegistrationTimeout { get; set; } = 15000;
 
     /// <summary>
+    /// Flag that controlls whether MediatR will attempt to register handlers that containg generic type parameters.
+    /// </summary>
+    public bool RegisterGenericHandlers { get; set; } = true;
+
+    /// <summary>
     /// Register various handlers from assembly containing given type
     /// </summary>
     /// <typeparam name="T">Type from assembly to scan</typeparam>
--- a/test/MediatR.Tests/GenericRequestHandlerTests.cs
+++ b/test/MediatR.Tests/GenericRequestHandlerTests.cs
@@ -1,0 +1,190 @@
+﻿using MediatR.Extensions.Microsoft.DependencyInjection.Tests;
+using Microsoft.Extensions.DependencyInjection;
+using Shouldly;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Emit;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using System.Reflection.PortableExecutable;
+using MediatR.Tests.MicrosoftExtensionsDI;
+
+namespace MediatR.Tests
+{
+    public class GenericRequestHandlerTests : BaseGenericRequestHandlerTests
+    {
+
+        [Theory]
+        [InlineData(9, 3, 3)]
+        [InlineData(10, 4, 4)]
+        [InlineData(1, 1, 1)]
+        [InlineData(50, 3, 3)]
+        public void ShouldResolveAllCombinationsOfGenericHandler(int numberOfClasses, int numberOfInterfaces, int numberOfTypeParameters)
+        {
+            var services = new ServiceCollection();
+
+            var dynamicAssembly = GenerateCombinationsTestAssembly(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);
+
+            services.AddMediatR(cfg =>
+            {
+                cfg.RegisterServicesFromAssemblies(dynamicAssembly);
+            });
+
+            var provider = services.BuildServiceProvider();
+
+            var dynamicRequestType = dynamicAssembly.GetType("DynamicRequest")!;
+
+            int expectedCombinations = CalculateTotalCombinations(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);
+
+            var testClasses = Enumerable.Range(1, numberOfClasses)
+                .Select(i => dynamicAssembly.GetType($"TestClass{i}")!)
+                .ToArray();
+
+            var combinations = GenerateCombinations(testClasses, numberOfInterfaces);          
+
+            foreach (var combination in combinations)
+            {
+                var concreteRequestType = dynamicRequestType.MakeGenericType(combination);
+                var requestHandlerInterface = typeof(IRequestHandler<>).MakeGenericType(concreteRequestType);
+
+                var handler = provider.GetService(requestHandlerInterface);
+                handler.ShouldNotBeNull($"Handler for {concreteRequestType} should not be null");
+            }            
+        }
+
+        [Theory]
+        [InlineData(9, 3, 3)]
+        [InlineData(10, 4, 4)]
+        [InlineData(1, 1, 1)]
+        [InlineData(50, 3, 3)]
+        public void ShouldRegisterTheCorrectAmountOfHandlers(int numberOfClasses, int numberOfInterfaces, int numberOfTypeParameters)
+        {  
+            var dynamicAssembly = GenerateCombinationsTestAssembly(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);          
+            int expectedCombinations = CalculateTotalCombinations(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);
+            var testClasses = Enumerable.Range(1, numberOfClasses)
+               .Select(i => dynamicAssembly.GetType($"TestClass{i}")!)
+               .ToArray();
+            var combinations = GenerateCombinations(testClasses, numberOfInterfaces);
+            combinations.Count.ShouldBe(expectedCombinations, $"Should have tested all {expectedCombinations} combinations");
+        }
+
+        [Theory]
+        [InlineData(9, 3, 3)]
+        [InlineData(10, 4, 4)]
+        [InlineData(1, 1, 1)]
+        [InlineData(50, 3, 3)]
+        public void ShouldNotRegisterDuplicateHandlers(int numberOfClasses, int numberOfInterfaces, int numberOfTypeParameters)
+        {
+            var dynamicAssembly = GenerateCombinationsTestAssembly(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);
+            int expectedCombinations = CalculateTotalCombinations(numberOfClasses, numberOfInterfaces, numberOfTypeParameters);
+            var testClasses = Enumerable.Range(1, numberOfClasses)
+               .Select(i => dynamicAssembly.GetType($"TestClass{i}")!)
+               .ToArray();
+            var combinations = GenerateCombinations(testClasses, numberOfInterfaces);
+            var hasDuplicates = combinations
+              .Select(x => string.Join(", ", x.Select(y => y.Name)))
+              .GroupBy(x => x)
+              .Any(g => g.Count() > 1);
+
+            hasDuplicates.ShouldBeFalse();
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenRegisterningHandlersWithNoConstraints()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateMissingConstraintsAssembly();
+
+            Should.Throw<ArgumentException>(() =>
+            {
+                services.AddMediatR(cfg =>
+                {
+                    cfg.RegisterServicesFromAssembly(assembly);
+                });
+            })
+            .Message.ShouldContain("When registering generic requests with more than two type parameters, each type parameter must have at least one constraint of type interface or class.");
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenTypesClosingExceedsMaximum()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateTypesClosingExceedsMaximumAssembly();
+
+            Should.Throw<ArgumentException>(() =>
+            {
+                services.AddMediatR(cfg =>
+                {
+                    cfg.RegisterServicesFromAssembly(assembly);
+                });
+            })
+            .Message.ShouldContain("One of the generic type parameter's count of types that can close exceeds the maximum length allowed");
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenGenericHandlerRegistrationsExceedsMaximum()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateHandlerRegistrationsExceedsMaximumAssembly();
+
+            Should.Throw<ArgumentException>(() =>
+            {
+                services.AddMediatR(cfg =>
+                {
+                    cfg.RegisterServicesFromAssembly(assembly);
+                });
+            })
+            .Message.ShouldContain("The total number of generic type registrations exceeds the maximum allowed");
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenGenericTypeParametersExceedsMaximum()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateGenericTypeParametersExceedsMaximumAssembly();
+
+            Should.Throw<ArgumentException>(() =>
+            {
+                services.AddMediatR(cfg =>
+                {
+                    cfg.RegisterServicesFromAssembly(assembly);
+                });
+            })
+            .Message.ShouldContain("The number of generic type parameters exceeds the maximum allowed");
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenTimeoutOccurs()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateTimeoutOccursAssembly();
+
+            Should.Throw<TimeoutException>(() =>
+            {
+                services.AddMediatR(cfg =>
+                {
+                    cfg.MaxGenericTypeParameters = 0;
+                    cfg.MaxGenericTypeRegistrations = 0;
+                    cfg.MaxTypesClosing = 0;
+                    cfg.RegistrationTimeout = 1000;
+                    cfg.RegisterServicesFromAssembly(assembly);
+                });
+            })
+            .Message.ShouldBe("The generic handler registration process timed out.");
+        }
+    }
+}
--- a/test/MediatR.Tests/GenericRequestHandlerTests.cs
+++ b/test/MediatR.Tests/GenericRequestHandlerTests.cs
@@ -186,5 +186,35 @@ namespace MediatR.Tests
             })
             .Message.ShouldBe("The generic handler registration process timed out.");
         }
+
+        [Fact]
+        public void ShouldNotRegisterGenericHandlersWhenOptingOut()
+        {
+            IServiceCollection services = new ServiceCollection();
+            services.AddSingleton(new Logger());
+
+            var assembly = GenerateOptOutAssembly();
+            services.AddMediatR(cfg =>
+            {
+                //opt out flag set
+                cfg.RegisterGenericHandlers = false;
+                cfg.RegisterServicesFromAssembly(assembly);
+            });
+
+            var provider = services.BuildServiceProvider();
+            var testClasses = Enumerable.Range(1, 2)
+                .Select(i => assembly.GetType($"TestClass{i}")!)
+                .ToArray();
+            var requestType = assembly.GetType("OptOutRequest")!;
+            var combinations = GenerateCombinations(testClasses, 2);
+
+            var concreteRequestType = requestType.MakeGenericType(combinations.First());
+            var requestHandlerInterface = typeof(IRequestHandler<>).MakeGenericType(concreteRequestType);
+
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs [hunk 1]
```diff
@@ -67,8 +67,28 @@ public class MediatRServiceConfiguration
...
+
+    /// <summary>
+    /// Configure the Maximum Amount of Generic RequestHandler Types MediatR will try to register.  To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int MaxGenericTypeRegistrations { get; set; } = 125000;
+
+    /// <summary>
+    /// Configure the Timeout in Milliseconds that the GenericHandler Registration Process will exit with error.  To Disable this constraint, set the value to 0.
+    /// </summary>
+    public int RegistrationTimeout { get; set; } = 15000;
```
hisuwh, line 90: 15 seconds seems like a pretty high default?  Are you expecting it to take a that long?

test/MediatR.Tests/GenericRequestHandlerTests.cs
```diff
@@ -0,0 +1,190 @@
...
+{
+    public class GenericRequestHandlerTests : BaseGenericRequestHandlerTests
+    {
+
+        [Theory]
+        [InlineData(9, 3, 3)]
+        [InlineData(10, 4, 4)]
+        [InlineData(1, 1, 1)]
+        [InlineData(50, 3, 3)]
+        public void ShouldResolveAllCombinationsOfGenericHandler(int numberOfClasses, int numberOfInterfaces, int numberOfTypeParameters)
```
hisuwh, line 26: What is this test doing?  This is not very readable

test/MediatR.Tests/GenericRequestHandlerTests.cs
```diff
@@ -0,0 +1,190 @@
...
+            var hasDuplicates = combinations
+              .Select(x => string.Join(", ", x.Select(y => y.Name)))
+              .GroupBy(x => x)
+              .Any(g => g.Count() > 1);
+
+            hasDuplicates.ShouldBeFalse();
+        }
+
+        [Fact]
+        public void ShouldThrowExceptionWhenRegisterningHandlersWithNoConstraints()
```
hisuwh, line 97: Why do you need to enforce this?

test/MediatR.Tests/SendTests.cs
```diff
@@ -183,4 +213,13 @@ public async Task Should_resolve_generic_void_handler()
...
     }
+
+    [Fact]
+    public async Task Should_resolve_multiple_type_parameter_generic_handler()
+    {
+        var request = new MultipleGenericTypeParameterRequest<TestClass1, TestClass2, TestClass3>();
+        await _mediator.Send(request);
+
+        _dependency.Called.ShouldBeTrue();
+    }
```
hisuwh, line 224: Can we also have a test on `Should_resolve_closed_handler_if_defined`

I.e. with something like `TestClass1PingHandler : IRequestHandler<VoidGenericPing<TestClass1>>`

So for a given generic type `T` if there is no closed handler but there is an open one it will resolve that.  But if there is an open one and a closed one it should resolve the closed one.

This is how my implementation works. This allows you to have a an OpenGeneric handler which works for most use cases but then override this with specific implementations if needed.

hisuwh (changes requested):

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: the test is taking as parameters a variable number of classes, interfaces, and type parameters.

The classes and interfaces are generated like so:

let's say that the number of classes is 3 and the number of interfaces is 2. The following classes and interfaces will be generated.

```csharp
//number of interfaces: 2
public interface ITestInterface1 {}
public interface ITestInterface2 {}

//number of classes: 3
//interface implementation is spread out evenly among the classes with rollover
public class TestClass1 : ITestInterface1 {}
public class TestClass2 : ITestInterface2 {}
public class TestClass3 : ITestInterface1 {} 
```
If the number of generic parameters is 2 then the following generic request / handler is made

```csharp
public DynamicRequest<T1, T2> : IRequest
    where T1 : ITestInterface1
    where T2 : ITestInterface2
{ }

public DynamicRequestHandler<T1, T2> : IRequestHandler<DynamicRequest<T1, T2>>
   where T1 : ITestInterface1
   where T2 : ITestInterface2
{
    public Task Handle(DynamicRequest<T1, T2> request, CancellationToken cancelationToken)
    {
         return Task.CompletedTask;
    }
}
```
The last thing the test does is generate all combinations for concrete DynamicRequest implementations. It then loops through each combination and creates a test that attempts to retrieve the service and ensure that it is registerd (not null);

```csharp
provider.GetService<IRequestHandler<TestClass1, TestClass2>>().ShouldNotBeNull();
provider.GetService<IRequestHandler<TestClass3, TestClass2>>().ShouldNotBeNull();
```

src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
zachpainter77: no I don't have any expectations.  I'm really open to whatever the team thinks is acceptable.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: because if no constraint is present then the the registration will try to register all types from the assembly when it isn't necessary. Type constraints should be present to limit the types that can close this generic type.

There isn't any way of knowing how many types are referenced by an assembly and MediatR will try and use to close the request handler types.

test/MediatR.Tests/SendTests.cs
zachpainter77: Yes I can make a test for that.. No problem.

zachpainter77: But that test might be redundant since the new mediatR registration removes the need to manually register them. And the manual registration becomes a duplicate service on the container.

src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
zachpainter77: There's no way to have expectations because the user's implementation is impossible to predict.

test/MediatR.Tests/SendTests.cs
zachpainter77: I'll have to add some type evaluator that tells mediator to ignore the test class used in a manual registration. Then I can test if that registration is there.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: If the consensus is that this shouldn't be enforced then I'm ok with that. Just say the word and I'll change it.

test/MediatR.Tests/SendTests.cs
zachpainter77: Actually what will determine if a manually registered closed handler is called over a generic one will be the order in which the registrations happen.

This is because the closed `IRequest<VoidGenericPing<Pong>` service is already registered when calling IMediatR.  And if you wish to override this service registration then you will need to add your manual registration after the call to `AddMediatR`.

I added this test.. I modified the `Dependency` class to add a property called `CalledSpecific`. The only handler that modifies the value is the newly added closed concrete implementation that you defined above: 
`TestClass1PingHandler : IRequestHandler<VoidGenericPing<Pong>>`

```csharp
 [Fact]
 public async Task Should_resolve_closed_handler_if_defined()
 {
     var dependency = new Dependency();
     var services = new ServiceCollection();
     services.AddSingleton(dependency);       
     services.AddMediatR(cfg => cfg.RegisterServicesFromAssemblies(Assembly.GetExecutingAssembly()));
     services.AddTransient(typeof(IRequestHandler<VoidGenericPing<Pong>>), typeof(TestClass1PingRequestHandler));
     var serviceProvider = services.BuildServiceProvider();
     var mediator = serviceProvider.GetService<IMediator>()!;

     var request = new VoidGenericPing<Pong>();
     await mediator.Send(request);

     dependency.CalledSpecific.ShouldBeTrue();
 }

```

As you can see the manual registration of the handler follows the call to `AddMediatR`. If the manual registration is moved above the call to `AddMediatR` then the test will fail because Microsoft DI uses the last registered implementation for a given service.  

Does this meet requirements?

zachpainter77: Furthermore I added this test that ensures the opposite (that the open request handler is resolved when a manual registration for request/handler is not defined but a manual registration exists)

```csharp
[Fact]
public async Task Should_resolve_open_handler_if_not_defined()
{
    var dependency = new Dependency();
    var services = new ServiceCollection();
    services.AddSingleton(dependency);
    services.AddMediatR(cfg => cfg.RegisterServicesFromAssemblies(Assembly.GetExecutingAssembly()));
    services.AddTransient<IRequestHandler<VoidGenericPing<PongExtension>>, TestClass1PingRequestHandler>();
    var serviceProvider = services.BuildServiceProvider();
    var mediator = serviceProvider.GetService<IMediator>()!;

    var request = new VoidGenericPing<Pong>();
    await mediator.Send(request);

    dependency.Called.ShouldBeTrue();
    dependency.CalledSpecific.ShouldBeFalse();
}
```

test/MediatR.Tests/GenericRequestHandlerTests.cs
hisuwh: > the registration will try to register all types from the assembly when it isn't necessary

Do I take it then that if you have `MyGenericRequest<T> where T : class` and `MyGenericRequestHandler<T>` then this will register an instance of `MyGenericRequestHandler<T>` for every class in the assembly?  And what happens if the type T is not in the same assembly as `MyGenericRequestHandler`?

test/MediatR.Tests/SendTests.cs
hisuwh: No that wouldn't meet my requirements because I would now have to manually register all closed handlers where this is handled automatically currently.

My implementation uses a function of my DI Container that allows registering types at runtime where they are missing (using this: https://jasperfx.github.io/lamar/documentation/ioc/registration/policies/)

So if a type is closed this is picked up by AddMediatR and takes precedence.  If I fire off a generic request without a closed handler the policy fires and resolves to the matching GenericHandler and closes this types.  

This adds a tiny overhead at runtime but only the first time this particular handler is needed and saves registering all possible closures (even if they are never called - which is what I believe this is doing).

Understanding more now how this works I think we need a way to opt out of this functionality as it is going to be hard to make this compatible with my solution, and I'm not sure this will work as well.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: I'm not sure if I understand your question.. I don't know what you mean by "register an instance". As far as I understand, registering a service is nothing more than an entry in a dictionary, more or less, of a keyed "service type" and a value "implementation type".  no instances are created unless you call `GetService<TService>()` which is what gets called behind the scenes most likely when you stuff a service in a constructor in .net core.

So no, an instance of `MyGenericRequestHandler<T>` is not created for each class in the assemblies.  But a registration entry is registered in the service collection for every single type that closes `MyGenericRequestHandler<T>`..  It is doing the exact same thing you are doing by manually creating closed registrations for your generic handlers.  It is just doing it by scanning the assembly and figuring out which classes close that type. and then registering them. In my opinion if you have have to manually create concrete handlers for each and every generic implementation, then it kind of defeats the purpose. 

Which is why I believe that checking for type constraints is, in my opinion, necessary.

test/MediatR.Tests/SendTests.cs
zachpainter77: yeah it's not hard to opt out. you can simply use the type evaluator. 

I'm pretty sure @jbogard doesn't want to use any third party DI containers such as lamar.

zachpainter77: I'm not really understanding your specific needs.. All I know is that this will resolve the issues of breaking changes for people who have manually registered generic handlers with more than one type parameter.

zachpainter77: Another possible way is to create a boolean configuration property: 

```csharp
public bool RegisterGenericHandlers { get; set; } = true;
```
then you could turn this off when calling `AddMediatR`:

```csharp
services.AddMediatR(cfg => 
{
    cfg.RegisterGenericHandlers = false;
    cfg.RegisterHandlersFromAssembliesContaining<Ping>();
});
```

zachpainter77: > Understanding more now how this works I think we need a way to opt out of this functionality as it is going to be hard to make this compatible with my solution, and I'm not sure this will work as well.

If anything this will just remove your dependency on lamar IoC.

hisuwh: I'm not suggesting using Lamar and I'm using Lamar for other reasons.

> this will resolve the issues of breaking changes for people who have manually registered generic handlers with more than one type parameter

I'm not manually registering generic handlers though. that's my point.

I think adding this configuration now would be a good idea
```
public bool RegisterGenericHandlers { get; set; } = true;
```

test/MediatR.Tests/GenericRequestHandlerTests.cs
hisuwh: Sorry I misspoke, I don't mean instances.  

My point is that type constraints might result in too many types being registered.  Probably worth making it clearer in the wiki how this works with type constraints so people can define suitably specific constraints.

Also to my other question how does the reflection handle looking up types?

So if I had something like this:

```
Domain.csproj
- MyGenericRequest<TEntity, TDto>
- MyGenericHandler<TEntity, TDto>
- MyType/MyTypeDto

Entities.csproj
- MyType
```

```cs
services.AddMediatR(cfg => 
{
    cfg.RegisterHandlersFromAssemblyContaining<MyTypeDto>();
});

services.GetInstance<IMediator>().Send(new MyGenericRequest<MyType, MyTypeDto>());
```

Will this work?

test/MediatR.Tests/SendTests.cs
crnd: > I think adding this configuration now would be a good idea
> 
> ```
> public bool RegisterGenericHandlers { get; set; } = true;
> ```

Is there any chance that existing applications will still break when `RegisterGenericHandlers` is set to `true`? If that is the case, I would say that this should be set to `false` by default. That way existing applications would function exactly the same way they've functioned this far and if someone wants to turn on the generics support, it would be a conscious choice. Of course this is a bit late now as `12.3.0` has already been released anyway.

Otherwise the major version of the package should be updated to 13 if the feature is on by default and it might cause issues with existing implementations.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: The code in the pr is set to throw an exception if you have a generic request handler that you are trying to register that 1) has more than two type parameters **AND** 2) does not have a type constraint defined.  This is to prevent too many registrations from happening.

from: ServiceRegistrar.cs

```csharp
 if (constraintsForEachParameter.Count > 2 && constraintsForEachParameter.Any(constraints => !constraints.Where(x => x.IsInterface || x.IsClass).Any()))
     throw new ArgumentException($"Error registering the generic handler type: {openRequestHandlerImplementation.FullName}. When registering generic requests with more than two type parameters, each type parameter must have at least one constraint of type interface or class.");
```

So, yes that example will work.

zachpainter77: If you intend on being this critical then I suggest that you clone the repository and test it yourself. Then you can be sure what works and what doesn't.  I'm not really trying to help you make your specific solution work as much as just fixing the issues mentioned in the PR.

test/MediatR.Tests/SendTests.cs
zachpainter77: > I'm not suggesting using Lamar and I'm using Lamar for other reasons.
> 
> > this will resolve the issues of breaking changes for people who have manually registered generic handlers with more than one type parameter
> 
> I'm not manually registering generic handlers though. that's my point.
> 
> I think adding this configuration now would be a good idea
> 
> ```
> public bool RegisterGenericHandlers { get; set; } = true;
> ```
you are most likely using lamar to register your generic handlers by scanning at runtime. That's just a guess. I don't know for sure.. but in any case:

This solution fixes the issues.. 

- The breaking change was caused because people had generic handlers with more than one type parameter defined in the assembly.
- MediatR failed to register them because it was hardcoded to only register generics with 1 type parameter.
- The PR now supports registration of generic handlers with multiple type parameters.
- The PR resolves the issues. Does it not?

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: > Domain.csproj
> - MyGenericRequest<TEntity, TDto>
> - MyGenericHandler<TEntity, TDto>
> - MyType/MyTypeDto
> 
> Entities.csproj
> - MyType
> ```
> 
> ```cs
> services.AddMediatR(cfg => 
> {
>     cfg.RegisterHandlersFromAssemblyContaining<MyTypeDto>();
> });
> 
> services.GetInstance<IMediator>().Send(new MyGenericRequest<MyType, MyTypeDto>());
> ```
> 
> Will this work?

To make this work you would simply need to make sure that the assemblies are added to the `AddMediatR` call.
```csharp
builder.Services.AddMediatR(cfg => cfg.RegisterServicesFromAssemblies([typeof(MyTypeDto).Assembly, typeof(MyType).Assembly]);
```

test/MediatR.Tests/SendTests.cs
zachpainter77: > Understanding more now how this works I think we need a way to opt out of this functionality as it is going to be hard to make this compatible with my solution, and I'm not sure this will work as well.

I think you don't really understand what this implementation is doing or how it works. I think you are scared that your app will break and you are kind of panicked.  

Rest assured.. This functionality will not require you to manually register all of your closing types.  That is what this implementation is doing.  I've posted several examples of this. I cannot be any more clear.

zachpainter77: > > I think adding this configuration now would be a good idea
> > ```
> > public bool RegisterGenericHandlers { get; set; } = true;
> > ```
> 
> Is there any chance that existing applications will still break when `RegisterGenericHandlers` is set to `true`? If that is the case, I would say that this should be set to `false` by default. That way existing applications would function exactly the same way they've functioned this far and if someone wants to turn on the generics support, it would be a conscious choice. Of course this is a bit late now as `12.3.0` has already been released anyway.
> 
> Otherwise the major version of the package should be updated to 13 if the feature is on by default and it might cause issues with existing implementations.

Also I prove that this won't break existing apps using mediator with the above tests:

```csharp
 [Fact]
 public async Task Should_resolve_closed_handler_if_defined()
 {
     var dependency = new Dependency();
     var services = new ServiceCollection();
     services.AddSingleton(dependency);       
     services.AddMediatR(cfg => cfg.RegisterServicesFromAssemblies(Assembly.GetExecutingAssembly()));
     services.AddTransient(typeof(IRequestHandler<VoidGenericPing<Pong>>), typeof(TestClass1PingRequestHandler));
     var serviceProvider = services.BuildServiceProvider();
     var mediator = serviceProvider.GetService<IMediator>()!;

     var request = new VoidGenericPing<Pong>();
     await mediator.Send(request);

     dependency.CalledSpecific.ShouldBeTrue();
 }
 ```

 1. If end users are not using generic handlers then they have nothing to worry about.
 2. if users did have generic handlers then they most likely had to manually register them or use a third party DI container like lamar, or autofac that will attempt to locate the closing types at runtime.  They too will not have any issues.
 
 The above test proves 2 to be accurate since mediatR registers all handlers correctly and then I manually register the handler.  Everything is working as expected and there are no errors.
 
 I'm confident that there doesn't need to be an opt in strategy and that an opt out strategy will suffice. I'm also confident that there doesn't need to be a major version increase.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: > Sorry I misspoke, I don't mean instances.
> 
> My point is that type constraints might result in too many types being registered. Probably worth making it clearer in the wiki how this works with type constraints so people can define suitably specific constraints.
> 

I don't understand how type constraints could do anything except reduce the number of services being registered. And in most cases explicitly target them.

src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
zachpainter77: I'm going to resolve this since there doesn't seem to be any issue with the default Timeout length.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: resolving since there is no suggestion being mentioned here only a clarification.

test/MediatR.Tests/SendTests.cs
zachpainter77: I've added an opt out functionality.. marking this as resolved.

test/MediatR.Tests/GenericRequestHandlerTests.cs
zachpainter77: I'm going to resolve this.. If you feel strongly that there should not be enforcement of type constraints on generic handlers with three or more type parameters then please unresolve this or start a new conversation.

jbogard (approved):

PR #1053, opened 2024-07-26
Validate arguments of public methods#CA1062

Zombach:
Source [CA1062](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062)

Cause
An externally visible method dereferences one of its reference arguments without verifying whether that argument is null (Nothing in Visual Basic).

You can [configure](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#configure-code-to-analyze) this rule to exclude certain types and parameters from analysis. You can also [indicate null-check validation methods](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062#null-check-validation-methods).

Rule description
All reference arguments that are passed to externally visible methods should be checked against null. If appropriate, throw an [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception) when the argument is null.

If a method can be called from an unknown assembly because it is declared public or protected, you should validate all parameters of the method. If the method is designed to be called only by known assemblies, mark the method internal and apply the [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute) attribute to the assembly that contains the method.

### DIFF
--- a/src/MediatR/Mediator.cs
+++ b/src/MediatR/Mediator.cs
@@ -35,8 +35,8 @@ public class Mediator : IMediator
     /// <param name="publisher">Notification publisher. Defaults to <see cref="ForeachAwaitPublisher"/>.</param>
     public Mediator(IServiceProvider serviceProvider, INotificationPublisher publisher)
     {
-        _serviceProvider = serviceProvider;
-        _publisher = publisher;
+        _serviceProvider = serviceProvider ?? throw new ArgumentNullException(nameof(serviceProvider));
+        _publisher = publisher ?? throw new ArgumentNullException(nameof(publisher));
     }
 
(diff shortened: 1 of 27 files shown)


zachpainter77:
I'm curious to see what happens with this PR. I have been thinking to myself of doing this with my code but never knowing for sure if the juice is worth the squeeze.

Zombach:
> I'm curious to see what happens with this PR. I have been thinking to myself of doing this with my code but never knowing for sure if the juice is worth the squeeze.

well I'm interested in that too

PR #1054, opened 2024-07-28
Fixed missing CancellationToken usage in MediatR class registration

samanazadi1996:
- Added missing CancellationToken parameter to ConnectImplementationsToTypesClosing method calls.

### DIFF
--- a/src/MediatR/Registration/ServiceRegistrar.cs
+++ b/src/MediatR/Registration/ServiceRegistrar.cs
@@ -46,15 +46,15 @@ public static class ServiceRegistrar
 
         ConnectImplementationsToTypesClosing(typeof(IRequestHandler<,>), services, assembliesToScan, false, configuration, cancellationToken);
         ConnectImplementationsToTypesClosing(typeof(IRequestHandler<>), services, assembliesToScan, false, configuration, cancellationToken);
-        ConnectImplementationsToTypesClosing(typeof(INotificationHandler<>), services, assembliesToScan, true, configuration);
-        ConnectImplementationsToTypesClosing(typeof(IStreamRequestHandler<,>), services, assembliesToScan, false, configuration);
-        ConnectImplementationsToTypesClosing(typeof(IRequestExceptionHandler<,,>), services, assembliesToScan, true, configuration);
-        ConnectImplementationsToTypesClosing(typeof(IRequestExceptionAction<,>), services, assembliesToScan, true, configuration);
+        ConnectImplementationsToTypesClosing(typeof(INotificationHandler<>), services, assembliesToScan, true, configuration, cancellationToken);
+        ConnectImplementationsToTypesClosing(typeof(IStreamRequestHandler<,>), services, assembliesToScan, false, configuration, cancellationToken);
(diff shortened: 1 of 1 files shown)


zachpainter77:
Just curious.. what does adding the cancellation tokens to the calls change?

samanazadi1996:
@zachpainter77 
Currently, the 'AddMediatR'  method allows users to customize timeout and configuration settings for MediatR. When users specify the 'RegistrationTimeout', it is expected that the relevant methods will adhere to these settings to manage system performance effectively and handle cancellation requests as needed.

zachpainter77:
> @zachpainter77 Currently, the 'AddMediatR' method allows users to customize timeout and configuration settings for MediatR. When users specify the 'RegistrationTimeout', it is expected that the relevant methods will adhere to these settings to manage system performance effectively and handle cancellation requests as needed.

The only relevant methods are the ones that already have the cancellation token passed.  Your PR doesn't do anything that is valuable.  The only thing that makes use of the timeout are registrations of open generic `IRequestHandler` implementations. Nothing else will ever make use of the timeout configuration.

samanazadi1996:
> > @zachpainter77 Currently, the 'AddMediatR' method allows users to customize timeout and configuration settings for MediatR. When users specify the 'RegistrationTimeout', it is expected that the relevant methods will adhere to these settings to manage system performance effectively and handle cancellation requests as needed.
> 
> The only relevant methods are the ones that already have the cancellation token passed. Your PR doesn't do anything that is valuable. The only thing that makes use of the timeout are registrations of open generic `IRequestHandler` implementations. Nothing else will ever make use of the timeout configuration.
> 
> There are only two methods that call `cancellationToken.ThrowIfCancellationRequested();` and those are `AddAllConcretionsThatClose` and `GenerateCombinations`. Those methods are only called when there are open generic implementations of `IRequestHandler`.
> 
> I know this because I authored the feature (#1048). I am the one who wrote the code to add the timeout configuration. It is only being used by the generics registration feature. If you are wishing to change this to work for all registrations globally then you will need to do more than pass the cancellation token to the method calls. You will need to add more places where the `cancellationToken.ThrowIfCancellationRequested();` is called as well.
> 
> An open generic example:
> 
> ```cs
> public class MyOpenGenericRequest<TEntity, TDto> : IRequest<TDto>
>     where TEntity : class, IEntity,
>     where TDto : class, IDto, new()
> {
>     public int EntityId { get; set; }
> }
> 
> public class MyOpenGenericRequestHandler<TEntity, TDto> : IRequestHandler<MyOpenGenericRequest<TEntity, TDto>, TDto>
>     where TEntity : class, IEntity,
>     where TDto : class IDto
> {
>     private IRepository<TEntity> _repo;
>     private IMapper _mapper;
> 
>     public MyOpenGenericRequestHandler(IRepository<TEntity> repo, IMapper mapper)
>     {
>         _repo = repo;
>         _mapper = mapper;
>     }
> 
>     public async Task<TDto> Handle(MyOpenGenericRequest<TEntity, TDto> request, CancellationToken cancellationToken)
>     {
>         var entity = await _repo.FindAsync(request.EntityId);
>         
>         return _mapper.Map<TDto>(entity);
>     }
> }   
> ```

Thank you for your explanation

Issue #1055, opened 2024-07-28
#1048 and v12.4.0 breaks RequestExceptionProcessorBehavior


luis-fss:
The example code in https://github.com/jbogard/MediatR/wiki#exception-handler-pipeline-step

```csharp
public class RequestExceptionProcessorBehavior<TRequest, TResponse, TException> : IRequestExceptionHandler<TRequest, TResponse, TException>
    where TRequest : notnull where TException : Exception
{
    public Task Handle(TRequest request, TException exception, RequestExceptionHandlerState<TResponse> state, CancellationToken cancellationToken)
    {
        throw new NotImplementedException();
    }
}
```

will throw an exception on startup: 

```System.ArgumentException: Error registering the generic handler type: PMCM.BlazorApp.Core.Pipeline.RequestExceptionProcessorBehavior`3. When registering generic requests with more than two type parameters, each type parameter must have at least one constraint of type interface or class.```

luis-fss:
Related:
#1047
#1048
#1038

zachpainter77:
I created a PR to remove the logic that causes that error. 

#1057

PR #1057, opened 2024-08-03
Make Auto Registration of Generic Handlers OPT-IN

zachpainter77:
- remove missing constraints logic.
- remove missing constraints tests.
- remove supporting method.
- make feature opt in via configuration

This is a direct fix for many people's issues regarding the enforcement of constraints for generic request handers having more than two generic type parameters. 

Direct fix for: #1055  , #1038

@jbogard  I have made this pr so that you can decide what you want to do. Either remove the feature completely or make it Opt-In until we can implement a better resolution strategy.

### DIFF
--- a/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
@@ -92,7 +92,7 @@ public class MediatRServiceConfiguration
     /// <summary>
     /// Flag that controlls whether MediatR will attempt to register handlers that containg generic type parameters.
     /// </summary>
-    public bool RegisterGenericHandlers { get; set; } = true;
+    public bool RegisterGenericHandlers { get; set; } = false;
 
     /// <summary>
     /// Register various handlers from assembly containing given type
--- a/src/MediatR/Registration/ServiceRegistrar.cs
+++ b/src/MediatR/Registration/ServiceRegistrar.cs
@@ -230,9 +230,6 @@ public static class ServiceRegistrar
             .Select(x => x.GetGenericParameterConstraints())
             .ToList();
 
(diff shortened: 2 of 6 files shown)


crnd:
Is this PR now just removing the protections from the registration process to make errors go away? Won't this now potentially cause users to end up having a lot of unwanted service registrations and potentially timeouts happening during application startup as the feature has been implemented opt-out instead of opt-in?

In my opinion the whole feature should either be removed or then the feature should be made opt-in to make sure that no one ends up getting this enabled by default by just upgrading to a newer minor version. As already stated before, this is anyway too late at this point as `12.3.0` and newer versions have already been released, but the current approach should have been a major version increment.

zachpainter77:
Yes it could result in too many registrations. I don't mind making it opt
in either.

zachpainter77:
> Is this PR now just removing the protections from the registration process to make errors go away? Won't this now potentially cause users to end up having a lot of unwanted service registrations and potentially timeouts happening during application startup as the feature has been implemented opt-out instead of opt-in?
> 
> In my opinion the whole feature should either be removed or then the feature should be made opt-in to make sure that no one ends up getting this enabled by default by just upgrading to a newer minor version. As already stated before, this is anyway too late at this point as `12.3.0` and newer versions have already been released, but the current approach should have been a major version increment.

Ok, I have made the feature Opt-In.  

image

zachpainter77:
@crnd 

I have made another PR alongside this one that will completely remove the feature.  So now it is only up to @jbogard to decide what PR to merge. 

- cheers

nhwilly:
This bit me pretty hard when I added `MassTransit` to my app.

alexkuznetsov:
@zachpainter77 @jbogard 

Hi, any chance of adding this and releasing the nuget package? 

After upgrading to version 12.4, we had a few interesting hours to investigate what the ... going on, 
and after some googling and reading issues on github we found a solution using 
the RegisterGenericHandlers configuration property.

jbogard:
Ugh so I think I'll push a version with this behavior opt-in but I want to tackle this much differently. Manually registering and scanning isn't really feasible.

PR #1060, opened 2024-08-04
Remove  Auto Registration of Generic Handlers Feature

zachpainter77:
This PR completely removes the Auto-Registration of open Generic Request Handlers.

@jbogard I have made this pr so that you can decide what you want to do. Either remove the feature completely or make it Opt-In until we can implement a better resolution strategy.

If this PR is accepted and merged then I will maintain this feature as an extension method in a separate nuget package.


jbogard:
I think what I'll do here is keep the tests because they represent the features I'd like to support, but remove the implementation. That it should be able to handle the use cases without any additional configuration. Is it possible to modify the PR to do that?

jbogard:
And yes the build will fail but that's OK

zachpainter77:
Sure I can do that.

zachpainter77:
@jbogard with the completion of pr https://github.com/jbogard/MediatR/pull/1057 do you think it's best to reject this pr and create a new one to remove the functionality but keep the failing tests to allow for development to tackle this in the way you were wanting to?

I feel like it would be easier to just pull in changes from the opt in pr and work from there.

zachpainter77:
Closing this PR and creating a new one.

PR #1068, opened 2024-09-09
Remove Generic Handler Support / Implementation (Leaving Failing Tests)

zachpainter77:
This PR removes the Generic Handler Support / Implementation and leaves the failing tests for future development / support behaviors as requested from @jbogard 

I can't say that I agree with this approach.  I feel like it would be better to leave the current implementation in while a better implementation is developed.  It doesn't really make sense to add the feature then remove it just to re-add it again in the future when someone can develop it in the way you are wanting.  I feel like it's better to just leave the working implementation in the opt in paradigm (opt in feature flag sort of behavior) and let people try to develop in a better way while the current feature is working.  That is more of a "Red, Green, Refactor" approach. It doesn't really make sense to me to remove a feature that :

1. isn't hurting anything (since it is now opt in) 
2. is currently meeting the test criteria that you want the feature to meet.

Those are my thoughts and opinions on it. However, this isn't my choice to make and that is fine.

Cheers.

### DIFF
--- a/src/MediatR.Contracts/MediatR.Contracts.csproj
+++ b/src/MediatR.Contracts/MediatR.Contracts.csproj
@@ -6,7 +6,6 @@
     <Description>Contracts package for requests, responses, and notifications</Description>
     <Copyright>Copyright Jimmy Bogard</Copyright>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <Nullable>enable</Nullable>
     <Features>strict</Features>
     <PackageTags>mediator;request;response;queries;commands;notifications</PackageTags>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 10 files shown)


jbogard:
Oh I can't merge this without replacing the new behavior, I was going to use this as a starting point for refactoring.

zachpainter77:
Yeah I think you could just use the opt-in branch as a starting point for the refactoring in my opinion.  Or we can rename this PR in some sort of way to let people try and get the tests passing with an implementation that satisfies your requirements. Idk.. lol

jbogard:
Well I don't think I can do this with the stock container. Lamar has support for this kind of thing but it replaces the stock container to do its thing. I'll just keep this existing behavior, even if I don't super like it lol.

PR #1081, opened 2024-10-22
pass missed cancellation tokens to ConnectImplementationsToTypesClosi…

Revazashvili:
pass missed cancellation tokens to ConnectImplementationsToTypesClosi…


remcoros:
LGTM (untested). might be nice to have a unit test for this

Revazashvili:
I believe the existing tests cover that, but the test project build fails due to vulnerabilities in some packages.

I installed different versions of System.Net.Http and System.Text.RegularExpressions packages explicitly. tests passed

PR #1093, opened 2025-01-17
Optimize Generic Handler Registration in MediatR

zachpainter77:
Ok @jbogard  I know how much you disliked the eager loading strategy of the first generic handler registration implementation so I took a shot at making a lazy loaded implementation that only scans the assembly when you request a generic handler service that is not already registered. This on demand approach seems to be exactly what you were describing to me. Please review. :)

## Summary
This PR introduces an optimized mechanism for registering generic handlers in `MediatR`. The current implementation scans all assemblies passed to `MediatR `during startup to find every possible concrete implementation and service type that satisfies all generic handler constraints. This PR modifies the behavior to scan and register generic handler services on-demand, triggered only when a specific service is requested.

This feature remains opt-in and can be enabled by setting the `RegisterGenericHandlers` configuration flag to true.

#### Optimized Generic Handler Registration:
-  The registration process now scans assemblies only when a specific service is requested, rather than eagerly scanning all possible types during startup.
-  Once a service is resolved, the registration is cached for future requests.

#### Dynamic Service Provider Integration:
- Introduced dynamic resolution and caching for generic handlers, minimizing the startup overhead.

#### Backward Compatibility:
- The feature remains optional and is controlled via the `RegisterGenericHandlers` flag in the `MediatR` configuration.
- All previous tests are passing.

#### Code Refactor:
- Removed old `ServiceRegistrar` logic used for previous eager loading of generic handlers. 
- Removed old configuration properties that attempted to restrain the user from soft locking their app.
- Extracted and modularized some key logic for resolving generic handler types.
- Improved readability and maintainability by reducing logic.

#### Improved Performance:
- Reduces startup time by avoiding eager scanning of all assemblies for generic handlers.
- Registration occurs only for the requested generic handler service, minimizing unnecessary work.

#### Lower Memory Overhead:
- Decreases the number of service descriptors stored in the `IServiceCollection `since only required services are registered.

#### Scalability:
- Particularly beneficial for applications with large assemblies and many potential generic handlers, as it avoids a full assembly scan.

#### Maintains Flexibility:
- Keeps the feature opt-in via `RegisterGenericHandlers`, ensuring no impact on existing applications unless explicitly enabled.

#### Compatibility:
- Fully backward-compatible with existing configurations and workflows.
- No third party dependencies required.

#### Potential Runtime Costs:
- The first request for a generic handler incurs a slight delay due to on-demand scanning and registration.
- Subsequent requests are unaffected due to caching.

#### Increased Complexity:
- The dynamic registration mechanism adds complexity to the codebase, which may require additional documentation and onboarding for contributors.

## Conclusion
This PR significantly optimizes the generic handler registration process, making it more efficient and scalable while maintaining backward compatibility. By deferring the registration of generic handlers to runtime and caching them for subsequent use, it strikes a balance between performance and flexibility.

### DIFF
--- a/test/MediatR.Tests/MicrosoftExtensionsDI/Handlers.cs
+++ b/test/MediatR.Tests/MicrosoftExtensionsDI/Handlers.cs
@@ -177,6 +177,10 @@ namespace MediatR.Extensions.Microsoft.DependencyInjection.Tests
 
     class MyCustomMediator : IMediator
     {
+        public MyCustomMediator() { }
+        public MyCustomMediator(IServiceProvider provider) { }
+        public MyCustomMediator(IServiceProvider provider, INotificationPublisher publisher) { }
+       
         public Task<object?> Send(object request, CancellationToken cancellationToken = new())
         {
             throw new System.NotImplementedException();
--- a/src/MediatR/MediatR.csproj
+++ b/src/MediatR/MediatR.csproj
@@ -34,6 +34,7 @@
     </PackageReference>
     <PackageReference Include="MediatR.Contracts" Version="[2.0.1, 3.0.0)" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="8.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
     <PackageReference Include="MinVer" Version="4.3.0" PrivateAssets="All" />
--- a/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatrServiceConfiguration.cs
@@ -69,27 +69,7 @@ public class MediatRServiceConfiguration
     /// Automatically register processors during assembly scanning
     /// </summary>
     public bool AutoRegisterRequestProcessors { get; set; }
-
-    /// <summary>
-    /// Configure the maximum number of type parameters that a generic request handler can have. To Disable this constraint, set the value to 0.
-    /// </summary>
-    public int MaxGenericTypeParameters { get; set; } = 10;
-
-    /// <summary>
-    /// Configure the maximum number of types that can close a generic request type parameter constraint.  To Disable this constraint, set the value to 0.
-    /// </summary>
-    public int MaxTypesClosing { get; set; } = 100;
-
-    /// <summary>
-    /// Configure the Maximum Amount of Generic RequestHandler Types MediatR will try to register.  To Disable this constraint, set the value to 0.
-    /// </summary>
-    public int MaxGenericTypeRegistrations { get; set; } = 125000;
-
-    /// <summary>
-    /// Configure the Timeout in Milliseconds that the GenericHandler Registration Process will exit with error.  To Disable this constraint, set the value to 0.
-    /// </summary>
-    public int RegistrationTimeout { get; set; } = 15000;
-
+        
     /// <summary>
     /// Flag that controlls whether MediatR will attempt to register handlers that containg generic type parameters.
     /// </summary>
--- a/src/MediatR/MicrosoftExtensionsDI/ServiceCollectionExtensions.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/ServiceCollectionExtensions.cs
@@ -47,11 +47,14 @@ public static class ServiceCollectionExtensions
             throw new ArgumentException("No assemblies found to scan. Supply at least one assembly to scan for handlers.");
         }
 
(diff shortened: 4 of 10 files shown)


zachpainter77:
> I'm just a random passerby who has used MediatR in the past, but I don't think this improvement is worth adding a whole new dependency to the core:
> 
> https://github.com/jbogard/MediatR/blob/761256f3b132ded5e5f6a2d33e3c092de6620f3a/src/MediatR/MediatR.csproj#L37
> 
> While it's fine to have the abstraction since the majority of people use it with ASP.NET Core anyway, there are some niche cases where people use MediatR with other DI containers (like SimpleInjector) and the app doesn't have the MSFT DI. In these cases, this MSFT dependency will be an unnecessary burden. Again, this isn't the end of the world, but it's not listed as a con. Also, this PR contains some commented-out code that shouldn't be in the final PR:
> 
> https://github.com/jbogard/MediatR/blob/761256f3b132ded5e5f6a2d33e3c092de6620f3a/src/MediatR/Registration/DynamicServiceProvider.cs#L71
> 
> There's also some dead code:
> 
> https://github.com/jbogard/MediatR/blob/761256f3b132ded5e5f6a2d33e3c092de6620f3a/src/MediatR/Registration/DynamicServiceProvider.cs#L15
> 
> This part also never can be `null`:
> 
> https://github.com/jbogard/MediatR/blob/761256f3b132ded5e5f6a2d33e3c092de6620f3a/src/MediatR/Registration/DynamicServiceProvider.cs#L95-L97
> 
> 
> etc

Thanks for your review! much appreciated!

I have fixed the following issues...

- removed property without reference.
- removed commented out code.
- removed redundant check for null open interface type.

I hear what you are saying with the dependency, however, the library is already dependent upon `Microsoft.Extensions.DependencyInjection.Abstractions` so I don't think this is a big issue for anyone who is already using this library. This library is necessary for MediatR to implement it's own custom ServiceProvider.  Since the default one cannot add services after it has already been built.  The requirement is to be able to add generic request / handler implementations at runtime after the default service provider has been built.  I'd imagine any other DI container that provides this functionality would also have this dependency on `Microsoft.Extensions.DependencyInjection`...

So I'll update the description / summary to include this as a con.  But I don't really see how it could that much of a burden for anyone currently using this with another container.

Again, Thank you for taking the time to review much appreciated. 

Cheers.

Rennasccenth:
Right on time, I'm looking forward this PR.

Once I finish my assessments and land a job(hope to), I wanna help with this.

@zachpainter77 theres anything that I can help you with?

zachpainter77:
Not at the moment but stay tuned I guess 😊

zachpainter77:
not stale.. This pr implements the on demand generic handler registration that @jbogard was looking for.

jbogard:
Sorry haven't forgotten about this one, just busy with day job

zachpainter77:
> Sorry haven't forgotten about this one, just busy with day job

No worries! understandable. :)

jbogard:
I think this is on the right track but still has some issues. The `Mediator` class uses static concurrent dictionaries to cache things, but in this implementation, nothing's actually ever cached. The transient constructor method in the (the call to `AddTransient`) is called every time `IMediator` is requested. You'll need to use a similar method used in the `Mediator` class to cache.

I'm not sure what issues would arise from the additional dependency. Maybe some extra types show up in IntelliSense?

zachpainter77:
> I think this is on the right track but still has some issues. The `Mediator` class uses static concurrent dictionaries to cache things, but in this implementation, nothing's actually ever cached. The transient constructor method in the (the call to `AddTransient`) is called every time `IMediator` is requested. You'll need to use a similar method used in the `Mediator` class to cache.
> 
> I'm not sure what issues would arise from the additional dependency. Maybe some extra types show up in IntelliSense?

Hmm.. I'm not quite sure I understand what you mean.. The implementation looks for a dependency and if it is not found using the default container it then checks the custom container. If the types are found it activates and resolves the dependency. If not then it adds it to the custom service provider and stores the registration for future use.  This in my understanding is cached.  So it only calls `AddTransient` the first time something is not found in either the default service provider or the custom one that registers dependencies during runtime.  The registration is stored as long as the application is running I suppose.  I'll run through the code to double check this.

zachpainter77:
Haven't had time to work on this quite yet. Keeping the pr alive with
comment.

luckyycode:
one thing I noticed is that if you got IMediator as transient/scoped and the DynamicServiceProvider gets created each time and does not cache

test/MediatR.Tests/MicrosoftExtensionsDI/Handlers.cs
```diff
@@ -176,7 +176,11 @@ class InternalPingHandler : IRequestHandler<InternalPing>
     }
 
     class MyCustomMediator : IMediator
-    {
+    {
+        public MyCustomMediator() { }
+        public MyCustomMediator(IServiceProvider provider) { }
+        public MyCustomMediator(IServiceProvider provider, INotificationPublisher publisher) { }
```
Arnab-Developer, line 182: You have added parameters (`provider` and `publisher`) in the constructor, but those parameters don't have any usage! Why is it needed?

Arnab-Developer (commented):
I was looking into the code changes and have one question which I have asked here. I am not too much familiar with the codebase, so may be my question makes no sense.

PR #1157, opened 2026-02-22
Fix: Cryptic "No constructor" Error When ILoggerFactory Is Not Registered (Issue #1153)

jbogard:
# Fix: Cryptic "No constructor" Error When ILoggerFactory Is Not Registered (Issue #1153)

## Context

After upgrading to MediatR 14.x, users get a confusing exception when resolving `IMediator`:

> `No constructor for type 'MediatR.Licensing.LicenseAccessor' can be instantiated using services from the service container and default values.`

**Why it happens:**

1. `ServiceRegistrar.AddRequiredServices` registers `LicenseAccessor` and `LicenseValidator` via bare generic `TryAddSingleton<T>()` (lines 473–474 in `ServiceRegistrar.cs`), with no factory lambda.
2. The DI container auto-constructs using the longest satisfiable constructor, but `ILoggerFactory` — required by both of `LicenseAccessor`'s constructors — is not registered by the application.
3. `serviceProvider.GetService<LicenseAccessor>()` in `CheckLicense` **throws** (not returns null) because the type is registered but its constructor cannot be satisfied.
4. The error names an internal type users have never heard of, with no hint they need to call `services.AddLogging()`.

## Files to Modify

- `src/MediatR/Registration/ServiceRegistrar.cs` — change how `LicenseAccessor` and `LicenseValidator` are registered (lines 473–474)
- `src/MediatR/MicrosoftExtensionsDI/MediatRServiceCollectionExtensions.cs` — remove now-dead fallback code in `CheckLicense` (lines 65–70)

### 1. Replace bare registrations with factory lambdas in `ServiceRegistrar.cs`

**Before (lines 473–474):**
```csharp
services.TryAddSingleton<LicenseAccessor>();
services.TryAddSingleton<LicenseValidator>();
```

**After:**
```csharp
services.TryAddSingleton<LicenseAccessor>(static sp =>
{
    var loggerFactory = sp.GetService<ILoggerFactory>()
        ?? throw new InvalidOperationException(
            "MediatR requires ILoggerFactory to be registered. " +
            "Call services.AddLogging() before services.AddMediatR().");
    var config = sp.GetService<MediatRServiceConfiguration>();
    return config != null
        ? new LicenseAccessor(config, loggerFactory)
        : new LicenseAccessor(loggerFactory);
});
services.TryAddSingleton<LicenseValidator>(static sp =>
{
    var loggerFactory = sp.GetService<ILoggerFactory>()
        ?? throw new InvalidOperationException(
            "MediatR requires ILoggerFactory to be registered. " +
            "Call services.AddLogging() before services.AddMediatR().");
    return new LicenseValidator(loggerFactory);
});
```

This produces a clear, actionable error at `IMediator` resolution time instead of the internal-type-naming cryptic message.

### 2. Simplify `CheckLicense` in `MediatRServiceCollectionExtensions.cs`

The `?? new LicenseAccessor(...)` and `?? new LicenseValidator(...)` fallbacks are now dead code: both types are always registered (via `TryAddSingleton`), so `GetService<T>()` will never return null — it either succeeds or throws. Simplify to `GetRequiredService` calls.

**Before (lines 65–70):**
```csharp
var licenseAccessor = serviceProvider.GetService<LicenseAccessor>() ?? new LicenseAccessor(
    serviceProvider.GetRequiredService<MediatRServiceConfiguration>(),
    serviceProvider.GetRequiredService<ILoggerFactory>()
);
var licenseValidator = serviceProvider.GetService<LicenseValidator>()
                       ?? new LicenseValidator(serviceProvider.GetRequiredService<ILoggerFactory>());
```

**After:**
```csharp
var licenseAccessor = serviceProvider.GetRequiredService<LicenseAccessor>();
var licenseValidator = serviceProvider.GetRequiredService<LicenseValidator>();
```

## Verification

```bash
dotnet test test/MediatR.Tests/MediatR.Tests.csproj
```

All 166 existing tests pass (they already register `NullLoggerFactory.Instance`). The fix only changes behavior when `ILoggerFactory` is absent — producing a clear actionable message instead of the cryptic "No constructor" error.

### DIFF
--- a/src/MediatR/MicrosoftExtensionsDI/MediatRServiceCollectionExtensions.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatRServiceCollectionExtensions.cs
@@ -62,12 +62,8 @@ public static class MediatRServiceCollectionExtensions
     {
         if (LicenseChecked == false)
         {
-            var licenseAccessor = serviceProvider.GetService<LicenseAccessor>() ?? new LicenseAccessor(
-                serviceProvider.GetRequiredService<MediatRServiceConfiguration>(),
-                serviceProvider.GetRequiredService<ILoggerFactory>()
-            );
-            var licenseValidator = serviceProvider.GetService<LicenseValidator>()
-                                   ?? new LicenseValidator(serviceProvider.GetRequiredService<ILoggerFactory>());
+            var licenseAccessor = serviceProvider.GetRequiredService<LicenseAccessor>();
+            var licenseValidator = serviceProvider.GetRequiredService<LicenseValidator>();
             
             var license = licenseAccessor.Current;
             licenseValidator.Validate(license);
--- a/src/MediatR/MicrosoftExtensionsDI/MediatRServiceCollectionExtensions.cs
+++ b/src/MediatR/MicrosoftExtensionsDI/MediatRServiceCollectionExtensions.cs
@@ -66,7 +66,7 @@ public static class MediatRServiceCollectionExtensions
                 serviceProvider.GetRequiredService<MediatRServiceConfiguration>(),
                 serviceProvider.GetRequiredService<ILoggerFactory>()
             );
-            var licenseValidator = serviceProvider.GetService<LicenseValidator>() 
+            var licenseValidator = serviceProvider.GetService<LicenseValidator>()
                                    ?? new LicenseValidator(serviceProvider.GetRequiredService<ILoggerFactory>());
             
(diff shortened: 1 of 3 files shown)


cheng93:
this is a breaking change for anyone using `Mediator.LicenseKey = "<license key here>";` instead of `AddMediatr`

`LicenseAccessor` and `LicenseValidator` won't be registered for DI

